PR #22, opened 2017-11-12
Revert SDK docs import

UniversalSuperBox:
The SDK docs import is causing build issues on RTD. Sorry...


UniversalSuperBox:
We'll need to discuss how we can better serve this app development documentation since it causes build timeouts. There's just too much of it for including in this tree...

Issue #3, opened 2017-08-09
Add landing pages


UniversalSuperBox:
Start from a good foundation and build from there, I always say.

Before we start writing a bunch of documents, we should get our landing pages in place. These include our indexes for top-level categories and each subcategory under them.

-   [ ]  About (About UBports, documentation structure)
-   [x]  User Guide
-   [x]    Installation
-   [x]   Daily Use
-   [x]   Hacking (Fancy command line stuff, things that require writable root)
-   [x]  Contribution
-   [x]   Bug Reporting
-   [x]   App Development
-   [ ]   OS Development
-   [x]   Translation
-   [x]   Documentation
-   [x]   Monetary Support

See how these all fit together (since I can't nest them on GitHub) [here](https://forums.ubports.com/topic/467/i-need-help-reorganizing-the-wiki).

doniks:
Just to be clear. Some of these pages will be *landing pages* like Daily Use that introduce the topic and then link to a bunch of other pages where the actual content is. 

Some other pages will not only be landing pages but will just hold the entirety of content on that topic: About, Monetary Support. But possible even Installation, Bug reporting, Translation and Documentation. 

Agree?

UniversalSuperBox:
Yes, that is correct.

doniks:
I just checked and it seems 
* Contribution is there
* Bug reporting is there 
* App development is there 
* Translation is there

doniks:
@UniversalSuperBox 
* **Translation:** This is done ain't it? What do you miss?
* **About**: Shouldn't we simply move https://docs.ubports.com/en/latest/about/introduction.html one level up? I can't think of anything else to say.
* **OS Development:** Do you think it's good enough to introduce and link to Halium? Or should we actually put something like this https://wiki.ubports.com/wiki/UBports-Development-Information in place as well?

NeoTheThird:
> About: Shouldn't we simply move https://docs.ubports.com/en/latest/about/introduction.html one level up? I can't think of anything else to say.

I think this is a good idea, or did you have something else in mind for that, @UniversalSuperBox? Maybe we could also put our process definitions in about, instead of contribute, with subsections for the HAL-, Middleware-, UI-, Marketing-, Community-, Infrastructure- and QA-steering committees, as well as Board of Trustees, Board of Directors and Membership Committee.

[...]

PR #36, opened 2017-12-17
Move and update issue tracking guidelines

NeoTheThird:
As discussed in #3, i wanted to move the issue tracking guidelines to about. I did not create the sub-folders i talked about, because they might be impractical for processes that affect more than one team. For now it's probably ok to just dump all defined processes in /about/processes.

I also updated the issue tracking guidelines and split the release schedule into a separate file.

The old media files that I added in my first failed attempt on importing the contents of the old docs.ubuntu.com are now removed.


NeoTheThird:
Don't merge yet, i have some more changes planned.

NeoTheThird:
Ok, i probably shouldn't let this grow bigger... Ready to merge.

PR #37, opened 2017-12-17
Delete documentation-todo.rst

NeoTheThird:
It's maybe better to use github issues for this...

### DIFF
--- a/contribute/documentation-todo.rst
+++ b/contribute/documentation-todo.rst
@@ -1,5 +1,0 @@
-
-Documentation TODOs
--------------------
-
-.. todolist::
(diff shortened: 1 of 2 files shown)


UniversalSuperBox:
Okay. :(

Would you prefer the `..todo::` directive stayed in the documentation, or do you want all of them removed as well?

NeoTheThird:
Dunno, just saw that that file was empty... What's the upside of using the todo directives?

UniversalSuperBox:
They create an entry on https://docs.ubports.com/en/latest/contribute/documentation-todo.html, which will lead back to the page. Without `..todolist::`, they simply make an orange message like in https://docs.ubports.com/en/latest/porting/building-ubports-boot.html#index-0

UniversalSuperBox:
Wrong button -_-

NeoTheThird:
Hmm... If you wanna keep it, don't merge it.

NeoTheThird:
Ok, yeah, I see, it's probably fine. I see how it could be handy. But we should document when to use todos and when to use issues somewhere.

PR #38, opened 2017-12-18
Create build script

NeoTheThird:
I think it's a lot more handy to just call the script. Should i also update the readme and contribute/documentation build guides?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# check if virtualenv already exists
+if [ -d "$HOME/ubportsdocsenv" ]; then
+  echo "Build environment found"
+  . ~/ubportsdocsenv/bin/activate
+else
+  echo "No build environment found. Creating..."
+  # install prerequisites
+  sudo apt install python-pip
+  sudo pip install virtualenv
+  # create virtualenv
+  virtualenv ~/ubportsdocsenv
+  . ~/ubportsdocsenv/bin/activate
+  # install build tools
+  sudo pip install sphinx sphinx_rtd_theme recommonmark
+fi
+# build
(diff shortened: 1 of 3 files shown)


NeoTheThird:
The script could of course be more sophisticated, i was a little lazy.

build.sh
```diff
@@ -0,0 +1,20 @@
...
+else
+  echo "No build environment found. Creating..."
+  # install prerequisites
+  sudo apt install python-pip
+  sudo pip install virtualenv
+  # create virtualenv
+  virtualenv ~/ubportsdocsenv
+  . ~/ubportsdocsenv/bin/activate
+  # install build tools
+  sudo pip install sphinx sphinx_rtd_theme recommonmark
```
UniversalSuperBox: Luckily we don't need recommonmark any more. Should be changed in the Contribute doc, shame on me.

build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# check if virtualenv already exists
+if [ -d "$HOME/ubportsdocsenv" ]; then
+  echo "Build environment found"
+  . ~/ubportsdocsenv/bin/activate
+else
+  echo "No build environment found. Creating..."
```
UniversalSuperBox: Maybe say where we're putting the doc so that no one gets scared? Also consider adding some hashmarks or something so that our status echos stand out from regular command output

NeoTheThird:
Do you want me to change the guide to use the script?

UniversalSuperBox:
If you have the time, definitely!

NeoTheThird:
Ok, I'll do it in a new PR though, so people can already use this.

PR #40, opened 2017-12-18
Added platform category

NeoTheThird:
My Idea for the platform category is to collect "how-stuff-works" information here, while the other categories are more "how-to-do-stuff".

It might seem counter-intuitive to put Halium here rather than under porting, as well as UITK, clickable and click in app-dev, but i think it makes sense.

At the moment it's just links to external resources, but at a later date we could also create a platform folder with subfolders on other topics.


UniversalSuperBox:
I was hoping GitHub would let me add a commit message to that. I put the Porting information under Platform.

Issue #44, opened 2017-12-20
documentation structure proposal


lduboeuf:
hi,
found that `daily use` or `advance use` is not very accurate
my proposition for documentation structure: 

- ABOUT
  -   ...

- CONTRIBUTE
  - ...

- UBUNTU TOUCH INTERNALS
  - overview ( here we talk about OS architecture and Mobile specific things  )

- HOWTOS
  - Install Ubuntu Touch ( Here maybe find all things related to how install on supported device, howto port to new devices )
  - Switch release channels
  - Shell access via adb
  - Shell access via ssh
  - Screen Casting your UT device to your computer
  - Reverse tethering

    - ...
- APP DEVELOPPEMENT
  - Overview (intro confinement, link to ubuntu touch internals, click, languages, content hub, notifications, ... )
  - Build & Deploy
    - Install ( raw style )
    - Clickable
    - Ubuntu SDK IDE
  - Webapps 
    - ...
  - Native apps
    - ... 
  - Desktop X apps  (with libertine and non-official way)
  - External resources (tutorials, templates, blogs)
   -  ...

lduboeuf:
maybe not the right place for fast collaborative document work. Etherpad or something somewhere ?

NeoTheThird:
Not that different from the current arch, what exactly are you trying to solve?

NeoTheThird:
So you're shuffling the order of the sections in the sidebar around and use more subheadings in the sidebar in userguide?

NeoTheThird:
Maybe send a PR with your changes so we can see what exactly it is you want

lduboeuf:
right, documentation is not an easy task but definitely needed :)

PR #48, opened 2017-12-22
project folders refactoring refs #44

lduboeuf:
Here is a part of my proposition.
Porting and installing on the same topic.
remove `daily use` `advanced use` and put all of docs in `howtos`
did not work on app developpement structure yet


NeoTheThird:
Porting has nothing to do with installation, why would you document that together? Porting is about as difficult as development tasks come, while installation is a very easy process every user can (and has to) go through...

lduboeuf:
this is your point of view :) . the final step is to install, so for me the same.

lduboeuf:
anyway, don't mind if my PR not relevant.  maybe removing daily use and advanced use is ok

lduboeuf:
ok maybe not accurate what i did. just the term "plaform" was a little ambigous, i understand that there is a need to simple end user install procedure and for devs a guide for porting new device. maybe we can only keep removing `daily use and advanced use` in favor of ` howtos` ?

NeoTheThird:
In tech, [platform](https://en.wikipedia.org/wiki/Computing_platform) is an abstract term that describes to the environment of software and hardware. So my idea [when adding the platform category](https://github.com/ubports/docs.ubports.com/commit/3de4a9a5d15e8812c3a40fa0b13e2f6caf6a345b) was that it describes how the system is designed and works, while other sections are about how to work with that system. But due to some other changes, that got a little watered down, so we might have to revisit that...

UniversalSuperBox:
It's my fault that Porting is under Platform right now, it should be in its own section. I misunderstood a previous pull request.

I don't like the idea of putting all information for using Ubuntu Touch under one HOWTO section. I forsee having a regular user's guide with some handy tips in it, not a Wiki-style free-for-all.

lduboeuf:
alright, no problem for all that. still work to do though to make a newbie easy to start and understand all the picture. merry Christmas.

PR #50, opened 2017-12-30
Update ssh.rst

stefanomelchior:
In my experience the default behaviour is that ssh is not enabled at all, so that I usually execute that command to enable its connection. A similar command workes (android-gadget-tools mtp enable) when I can reach the mobile device with MTP, but it was not contextual in this page

### DIFF
--- a/userguide/advanceduse/ssh.rst
+++ b/userguide/advanceduse/ssh.rst
@@ -42,6 +42,10 @@ Now start the ssh server::
 To make sure the ssh server is automatically started in the future, execute::
 
     sudo setprop persist.service.ssh true
+    
+In case the ssh was not enabled before, use the following command to permit ssh connexion form outside the mobile device::
+
+    android-gadget-tools ssh enable
 
 Connect
 -------


doniks:
Are you sure that you did execute the aforementioned _sudo service ssh start_ ? In my experience that does the same thing. 

(Btw, connection is misspelled)

stefanomelchior:
https://askubuntu.com/questions/348714/how-can-i-access-my-ubuntu-phone-over-ssh

A long, but useful explanation of the advantages of using android-gadget-service.

Anyway it seems they are pretty similar, even thought a-g-s seems to be a bit more sure because it doesn’t require developer_mode and writable userdata enabled. It can be perform by phablet user.

Thanks for the pointing.
Bye

Stefano Melchior
Email: stefano.melchior@me.com -stefanomelchior@icloud.com
Mobile:  +34-627-006603
Skype: stefanomelchior
Twitter: @stefanomelchior

Sent from my iPhone 6+ - Enviado desde mi iPhone 6+

On 31 Dec 2017, at 01:14, doniks <notifications@github.com> wrote:

Are you sure that you did execute the aforementioned sudo service ssh start ? In my experience that does the same thing.

(Btw, connection is misspelled)

NeoTheThird:
@stefanomelchior just fyi, you might want to disable your email signature when answering to a github email. I removed your phone number from your comment for now.

stefanomelchior:
Yep, thanx Jan

Stefano Melchior

On 3 Jan 2018, at 13:36, Jan Sprinz <notifications@github.com> wrote:

@stefanomelchior just fyi, you might want to disable your email signature when answering to a github email. I removed your phone number from your comment for now.

PR #52, opened 2018-01-03
Remove ReCommonMark and add translation support

UniversalSuperBox:
This one looks scarier than it really is.

Removing ReCommonMark is something that I've wanted to do for a long time, this was just a good excuse to do it.

I've added an update-translations script to the repository (and languages.sh, which just contains the list of languages to output). I've also used the script to generate some starting po-files for translation.

Once that's done, we'll need to add this to Weblate using [these instructions](https://docs.weblate.org/en/latest/devel/starting.html?highlight=sphinx#translating-documentation-using-sphinx). Probably wouldn't be a bad idea to delete [this repository](https://github.com/ubports/docs.ubports.com-localization) as well.

Some considerations:

Once this is done, it will be a very not-good idea to move files from one folder to another, since I expect that sphinx will just delete the translations and add blank ones in the new folder.


NeoTheThird:
Lgtm!

> Once this is done, it will be a very not-good idea to move files from one folder to another, since I expect that sphinx will just delete the translations and add blank ones in the new folder.

Are you sure? If it uses gettext (which i think it does), is should be able to recognize changes in strings with no issue whatsoever.

PR #56, opened 2018-01-07
Tut for Webapp

Naturalbornhippy:
Today i tried to find a tutorial for creating webapps and i didnt find one. 
This is why i created this first page (needs still some more work, but it will help or?)


Naturalbornhippy:
Hm, can someone explain the error with travis to me?

PR #59, opened 2018-01-07
Webapps tutorial

Naturalbornhippy:
Hi again,

i figured out how to test this. sorry for the first three untested requests. I thought if it is displayed it is enough. Now i have build it on my machine and this build runs. Hopefully it does now on the travis main repo too. If not i will stop my work and someone has to tell me how to do this.

For the request: 
I didnt found a real tutorial for webapp development (some threads on the forums, but nothing in the docs or wikis). Thats why i wrote this small howto. I hope this works for everyone. i will improve the page from time to time.

### DIFF
--- a/appdev/webapp.rst
+++ b/appdev/webapp.rst
@@ -44,5 +44,5 @@ In this folder you will find a file called: namespace.app_name_app_version_all.c
 
 Upload it to OpenStore
 ---------
-This package can now be uploaded to  _OpenStore <https://open.uappexplorer.com/submit>_. Just create an accound an follow the instructions on the screen.
+This package can now be uploaded to  `OpenStore <https://open.uappexplorer.com/submit>`__. Just create an accound an follow the instructions on the screen.
 
--- a/appdev/webapp.rst
+++ b/appdev/webapp.rst
@@ -1,0 +1,49 @@
+Creating a Webapp
+===============
+Webapps are small applications, that load content in a browser. This tutorial for ubuntu will just create a starter for a webpage, that can be displayed easily in a mobile device.
+
+
+alternate-webapp-generator
+---------
+The alternate-webapp-generator will download all the stuff you need for a webapp and will add the the url, icon and a lot of meta-information to the app.
+In the end a .click package will be created, which can be uploaded to the OpenStore
+
+Download the following zip-file and unzip
+::
+    https://github.com/Winael/ubuntu/blob/master/ubuntuphone/webapps/Reboot/alternate-webapp-generator.zip
+
+Make the app-creator executable:
+
+::
+    chmod +x /path/to/alternate-webapp-generator.sh
+
+Now we can start and edit the config.cfg.
+::
+    # config.cfg
+
+    export namespace="namespace"
+
+    export app_name="yourname"
+    export app_title="YourTitle"
+    export app_url="https://ubports.com/"
+    export app_description="A webapp"
+    export app_UA="Mozilla/5.0 (Linux; Android 5.0; Nexus 5) AppleWebkit/537.36 (KHTML, like Gecko) Chrome/38.0.2125.102 Mobile Safari/537.36"
+    export app_version="0.1"
+
+    export maintainer_name="yourname"
+    export maintainer_email="your@email.com
+
+    export icon_path="/path/on/your/harddisk"
+
+Its important that the namespace and the app_name is in lowercase - if you add it in uppercase the build will fail.
+Lets build the webapp now
+::
+    ./alternate-webapp-generator.sh 
+
+This should now create a folder in the same directory which is named after the app_name attribute from the config.cfg
+In this folder you will find a file called: namespace.app_name_app_version_all.click
+
+Upload it to OpenStore
+---------
+This package can now be uploaded to  _`OpenStore <https://open.uappexplorer.com/submit>`_. Just create an accound an follow the instructions on the screen.
+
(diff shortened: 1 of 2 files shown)


UniversalSuperBox:
For future reference, if you push more commits to the branch that you select in your Pull Request, they'll automatically be added to the PR.

UniversalSuperBox:
image

How do you feel about this addition to the layout, @NeoTheThird ?

appdev/webapp.rst
```diff
@@ -0,0 +1,55 @@
...
+alternate-webapp-generator
+--------------------------
+
+The alternate-webapp-generator will download all the stuff you need for a webapp and will add the the url, icon and a lot of meta-information to the app.
+In the end a .click package will be created, which can be uploaded to the OpenStore
+
+Download the following zip-file and unzip
+::
+
+    https://github.com/Winael/ubuntu/blob/master/ubuntuphone/webapps/Reboot/alternate-webapp-generator.zip
```
NeoTheThird: Either make this a command, or a hyperlink. Just having the url isn't handy.

appdev/webapp.rst
```diff
@@ -0,0 +1,55 @@
...
+
+    chmod +x /path/to/alternate-webapp-generator.sh
+
+Now we can start and edit the config.cfg.
+::
+
+    # config.cfg
+
+    export namespace="namespace"
+
```
NeoTheThird: Do we want the whitespace here?

appdev/webapp.rst
```diff
@@ -0,0 +1,55 @@
...
+::
+
+    https://github.com/Winael/ubuntu/blob/master/ubuntuphone/webapps/Reboot/alternate-webapp-generator.zip
+
+Make the app-creator executable:
+::
+
+    chmod +x /path/to/alternate-webapp-generator.sh
+
+Now we can start and edit the config.cfg.
```
NeoTheThird: What is a config.cfg, something ducks walk on? Jk, but this comes a little out of the blue. Maybe explain above that in the process we'll create a new file that holds all the information, called config.cfg.

lduboeuf:
there is now `clickable init` that allow us to start with a webapp template, also `webapp creator` on openstore.

lduboeuf:
maybe have also a reference or a summary of this: https://docs.ubuntu.com/phone/en/web/ubuntu-webapps-guide ?

appdev/webapp.rst
```diff
@@ -0,0 +1,55 @@
...
+::
+
+    # config.cfg
+
+    export namespace="namespace"
+    
+    export app_name="yourname"
```
NeoTheThird: whycantthishavespaces? The namespace for the app-id is taken from the namespace field, so this should be formatted human readable "Your Name".

appdev/webapp.rst
```diff
@@ -0,0 +1,55 @@
...
+    export app_title="YourTitle"
```
NeoTheThird: This can have spaces as well

appdev/webapp.rst
```diff
@@ -0,0 +1,55 @@
...
+
+    ./alternate-webapp-generator.sh 
+
+This should now create a folder in the same directory which is named after the app_name attribute from the config.cfg
+In this folder you will find a file called: namespace.app_name_app_version_all.click
+
+Upload it to OpenStore
+----------------------
+
+This package can now be uploaded to  `OpenStore <https://open.uappexplorer.com/submit>`__. Just create an account an follow the instructions on the screen.
```
NeoTheThird: You can't create an account on the openstore, but log in with ubuntuone or github.

NeoTheThird:
Good points, @ldeboeuf! Using clickable would probably be a little cleaner, and it allows for cross-plattform instructions. This guide here will (i think) only work on 16.04.

Summing up the important parts of the document you linked to would also be a good idea, but of course without the ubuntu-sdk-ide stuff, since that is unsupported.

appdev/webapp.rst
Naturalbornhippy: Its included in the zip package - i will add a line with this information

appdev/webapp.rst
Naturalbornhippy: Its from the template file, but i can remove it from here - i dont mind

appdev/webapp.rst
Naturalbornhippy: Answer to the last two commands: actually both things would crash that script:
for example: 'ipes-ent.ipes ent' not properly formatted

appdev/webapp.rst
NeoTheThird: I meant the line breaks, actually, not the space. A space between the comment operator and the comment is generally considered a good practice, so i guess that should stay

appdev/webapp.rst
```diff
@@ -13,17 +13,17 @@ In the end a .click package will be created, which can be uploaded to the OpenSt
 Download the following zip-file and unzip
 ::
 
-    https://github.com/Winael/ubuntu/blob/master/ubuntuphone/webapps/Reboot/alternate-webapp-generator.zip
+    wget https://raw.githubusercontent.com/Winael/ubuntu/master/ubuntuphone/webapps/Reboot/alternate-webapp-generator.zip
 
 Make the app-creator executable:
 ::
 
     chmod +x /path/to/alternate-webapp-generator.sh
```
NeoTheThird: You could make this codeblock a little nicer if you also added the unzip command and joined them all into one block:
```
Download, extract and make the script executable::

    wget ...
    unzip...
    cd...
    chmod...

```

UniversalSuperBox: Also, while technically correct to do it this way, I'd prefer that the two colons were at the end of the previous line.

```
Download the following zip-file and unzip::

    code

Make the app-creator executable::

    code
```

PR #68, opened 2018-01-14
check-kernel-config: path

doniks:


UniversalSuperBox:
I've just made check-kernel-config executable in the ubports-boot repository. Does that fix this?

doniks:
did a repo sync and yes it's executable now. 

I think it adding the path from the repo root dir still makes it clearer to the user

PR #75, opened 2018-01-18
Update to ssh document

Acid-Override:
removing the setprop instructions as they are depreciated and replaced by android-gadget-service.


doniks:
Are we sure android-gadget-service deprecates this? 

Dont get me wrong, I don't have a strong opinion for either one. One sounds a bit more androidy the other more linuxy.

Do you have a reference for it deprecating the existing one or one of them causing a problem?

UniversalSuperBox:
The `service` command will run ssh for this boot. `android-gadget-service` will edit `ssh.override` so that ssh runs on every boot.

Issue #41, opened 2017-12-19
App developpement overview - quick start


lduboeuf:
Would be great to have a short overview.  
What languages we can use, short introduction of `click` packaging format, apparmor, confinement constrains.
Links to 'hello-world' templates (i think brian is going to add start examples, kind of a `clickable init` ).

lduboeuf:
WIP #80

PR #80, opened 2018-02-09
refs #41 add general informations for app developpment

lduboeuf:
don't know why my 4 last commits still here in the PR (already merged upstream). i did a merge upstream/master locally before


appdev/index.rst
```diff
@@ -6,16 +6,75 @@ Welcome to an open source and free platform under constant scrutiny and improvem
...
+        "policy_groups": [
+            "networking",
+            "webview",
+            "content_exchange"
+        ],
+        "policy_version": 1.3
+    }
+
+Non exhaustive policy groups:
+
```
lduboeuf: not sure if always available

appdev/index.rst
```diff
@@ -6,16 +6,75 @@ Welcome to an open source and free platform under constant scrutiny and improvem
...
 
-.. todo:
+Ubuntu Touch has three types of applications: Web Applications (WebApps), Scopes and Native Applications.
+Applications are packaged, distributed and deployed using a format called `click <https://click.readthedocs.io/en/latest/>`_ packaging.
+Languages of choice are QML or HTML5 for the UI and can be Javascript, Qt, C++, Python, Go for the logic.
+
+Security and app isolation
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+All Ubuntu apps and scopes are confined respecting `AppArmor <https://wiki.ubuntu.com/AppArmor>`_ access control mechanism, meaning they only have access to their own resources and are isolated from other apps and parts of the system. The developer must declare which policy groups are needed for the app or scope to function properly with a `manifest.json` file.
```
sverzegnassi: [QUESTION] Might it be better to link [Application Confinement](https://wiki.ubuntu.com/SecurityTeam/Specifications/ApplicationConfinement#App_confinement_with_AppArmor) from _Ubuntu Wiki_, which seems to provide more complete information?

appdev/index.rst
```diff
@@ -6,16 +6,75 @@ Welcome to an open source and free platform under constant scrutiny and improvem
...
+
+Example `manifest.json` file:
```
sverzegnassi: [NEEDS-FIX] AppArmor profile declaration is not part of `manifest.json`, but it's a requirement of the [default hooks](https://click.readthedocs.io/en/latest/hooks.html) available for Click packages.

Could you please add an actual example of `manifest.json` that links to an AppArmor profile (e.g. for an SDK app)?

appdev/index.rst
sverzegnassi: The list looks good to me. It only includes the public policies marked as "common", which is enough to get started with Ubuntu Touch development. Good choice!

lduboeuf:
can find any web site for `.desktop`  specifications on Ubtun Touch

sverzegnassi:
Looks good to me! Great job!

Desktop file spec is available at: https://standards.freedesktop.org/desktop-entry-spec/latest/

It's worth to note that not all the keys or values are equally supported on Ubuntu Touch, because of the particular confined environment. Also, there is a number of Ubuntu Touch extensions that only work with Unity 8.

I don't know if such differences have been tracked in some downstream project, we'd need to investigate further.
For the time being, we'd be better to keep the `.. todo:` entry.

I'm going to merge your PR as-is, since it's already a huge improvements.
Thanks again!

lduboeuf:
awesome, you're welcome :)

PR #83, opened 2018-02-12
userguide refactoring

lduboeuf:
this is a proposal restructuring for user guide section. As daily use or advanced used may not be accurate imo.
Or i was thinking of dividing this section to "common tasks" (adb, ssh, cal sync...) and "hacking" for the things that may break things.
For now i propose "HowTos" main subsection.


lduboeuf:
AS discussed on Telegram, i see more know the target audience of "daily use" and "advanced use". I close this PR, but maybe still need to move some articles to the right place.

PR #86, opened 2018-02-16
system software: add core apps link

lduboeuf:
Section "System Software" is confusing for me as i always heard about "core apps" or "system apps" but not "system software" or "system components". I did not modify titles though.

### DIFF
--- a/appdev/index.rst
+++ b/appdev/index.rst
@@ -173,7 +173,10 @@ A collection of external resources
 System Software
 ---------------
 
-Working on system components
+Ubuntu Touch image ship a set of applications also known as "core apps".
+See `core apps repositories on github <https://github.com/search?q=topic%3Acore-app+org%3Aubports&type=Repositories>`__
+
+This resource will help you get started with working on system components
 
 .. toctree::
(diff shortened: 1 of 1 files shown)


appdev/index.rst
```diff
@@ -173,7 +173,10 @@ A collection of external resources
 System Software
 ---------------
 
-Working on system components
+Ubuntu Touch image ship a set of applications also known as "core apps".
+See `core apps repositories on github <https://github.com/search?q=topic%3Acore-app+org%3Aubports&type=Repositories>`__
+
+This resource will help you get started with working on system components
```
UniversalSuperBox: It's good that this is being clarified. However, this doesn't really hit the goal of the page it links to. The article does list how to build core apps, but it's more about building any package installed on the Ubuntu Touch system. This includes libraries, system services, and other components that are preinstalled.

lduboeuf: ok, i also thought maybe it was more than core apps development.
That is fine i close this PR. Will see later then where is the best place to talk about "core apps"

PR #87, opened 2018-02-18
Desktop App Scope is in Open Store

jonnius:
Desktop App Scope is not to be installed from Canonical Store anymore


UniversalSuperBox:
Could you make "desktop app scope" a link to the application? https://open-store.io/app/libertine-scope.ubuntu

jonnius:
Good idea. Done.

PR #88, opened 2018-02-18
Made sentance less awkward.

Seshpenguin:


UniversalSuperBox:
I've tweaked that a bit. What do you think?

Seshpenguin:
Looks good

PR #90, opened 2018-02-20
add extensible folder structure for appdev

lduboeuf:
in order to be able to develop more about scopes, native apps and webapps, i propose to add this folder structure.
Also imported webapps guide from https://docs.ubuntu.com/phone/en/web/ubuntu-webapps-guide


lduboeuf:
humm maybe i  should put all webapps doc in a seperate file say "guide.rst" so that we keep index.rst as clean as possible and be able to extend with other informations ( refs PR #58 ) ?

PR #91, opened 2018-02-21
webapp: import webapp debugging from wiki - add  some external links

lduboeuf:


appdev/webapp/webdebug.rst
```diff
@@ -0,0 +1,31 @@
+Debug web application
+=====================
+
+Developers of a web application will probably do most of their coding
+and debugging in the usual desktop browser environment. Because UBports'
+browser has excellent compliance with modern web standards, such code
+will most often operate on the UBports browser without further changes.
```
NeoTheThird: maybe go with a little less self-adulation

lduboeuf: ok, can you modify it directly?, i've just copied-paste that from wiki :)

PR #136, opened 2018-08-28
Hunt down the typos

nikalon:
I ran `aspell` through the entire repository to find and fix most of the noticeable typos I came across. I don't think I fixed all of them, but it's cleaner now :)


nikalon:
Yeah, I only edited manually the .rst files and the pots were generated by `update-translations.sh` script as per instructions in https://docs.ubports.com/en/latest/contribute/documentation.html

PR #142, opened 2018-09-22
Support macOS to build the project

tatarbj:
Currently as nproc is used in build.sh, it fails as nproc doesn't exist in mac. In my PR I extended build.sh for support mac and linux with checking the uname and based on that decide usage of nproc or sysctl.


UniversalSuperBox:
That covers most of our bases. Could it be that we would also want to support the more common other UNIXes, like FreeBSD? If so, I think we could do 

```
if linux:
    nproc
else:
    sysctl -n hw.ncpu
```

It seems that at least some of the BSDs should also work this way.

UniversalSuperBox:
@tatarbj, are you available to make that change to the `if` statement?

tatarbj:
Hi @UniversalSuperBox, sorry for the late reply, somehow i missed your comment. I've changed the way how that if statement identifies which way should be used, but having only a suse and my own mac on my side to test, i'm not fully sure it's what you've meant. At least tests are green, but it would be better to get your opinion on it :)

UniversalSuperBox:
As far as I can see (and test), this is ready to be merged! I'll come back to it after GitHub is not broken and the checks run...

PR #143, opened 2018-09-25
Revamped the appdev pages

bhdouglass:


appdev/index.rst
```diff
@@ -3,190 +3,103 @@ App development
...
-  - X-Ubuntu-Touch: ``true`` to make the app visible
-  - X-Ubuntu-XMir-Enable: ``true`` if your app is built for X
+`Clickable <http://clickable.bhdouglass.com/en/latest/>`_ is a meta-build system for Ubuntu Touch applications that allows you to compile, build, test and publish `click` packages and provides various templates to get you started with app development.
+It is currently the easiest and most convenient way of building click packages for Ubuntu Touch.
+You can use any code editor or IDE that you choose and build your apps from the commandline with Clickable.
 
-.. todo:
-  link to official .desktop specifications
+Alternatively there is the old `Ubuntu SDK IDE <https://docs.ubuntu.com/phone/en/platform/sdk>`__.
+But it is no longer supported by Canonical, and UBports has chosen to not support it either due to lack of manpower.
```
UniversalSuperBox: Consider rewording this so the sentence doesn't need to start with "But"

bhdouglass: Even though English is my first language I still suck at it. But I can see about rewriting that 🙂

bhdouglass:
Yeah it's mostly just habit to keep line length down. Will it be a problem for anything? I can change it.

UniversalSuperBox:
In [another pending PR](https://github.com/ubports/docs.ubports.com/pull/141/files#diff-c58178921b94dbbb599fed5dd5a39012R113) I added a "line length" section to the style guide. Its new instruction is "don't break lines arbitrarily. Use your editor's word wrap". It makes it a little easier to update the documentation without worrying about needing to reflow lines. Your method definitely keeps the line lengths down *and* doesn't cause problems with reflowing. Hmm, maybe that should be the new standard.

What do you think?

bhdouglass:
Either way is fine by me (as long as I can get my editor to word wrap only for rst files and not everything else).

appdev/index.rst
wayneoutthere: should writers group be engaged in grammar stuff?  And, does this comment right here screw up the project?  We shall see when I press 'add single comment'...

bhdouglass: Including the writers group on this stuff makes sense to me, I can do technical stuff but not word stuff :+1:

bhdouglass:
@UniversalSuperBox done!

PR #150, opened 2018-10-22
Add information on preinstalled apps

UniversalSuperBox:
We previously haven't had a canonical list of core apps anywhere. We do now.


bhdouglass:
@UniversalSuperBox should we include Telegram, Doc Viewer, and Sudoku somewhere? They aren't preinstalled but they are "core apps".

UniversalSuperBox:
How about a new heading, "Apps maintained by UBports developers"?

bhdouglass:
Sounds good

PR #151, opened 2018-10-22
Update system software documentation

UniversalSuperBox:
This PR changes the "App development" chapter to "Development" and adds a new System software development section underneath it. It also updates the system software development documentation.

PR #154, opened 2018-10-31
Developing system-software: running tests with crossbuilder

mardy:
Add information on how to develop locally with crossbuilder, and be able to run unit tests.


UniversalSuperBox:
As discussed in chat, this is partially superceded by #151. Also, these abilities of Crossbuilder are probably better broken out as a large guide rather than a small aside in a different article.

PR #177, opened 2019-01-08
 Update build and translation workflow

UniversalSuperBox:
Weblate was not using the .pot file to automatically update translation sources. For this reason, we will need to continue rebuilding them as we previously did, with sphinx-intl update...

This comes with its own challenges (merge conflicts everywhere on Weblate) but should unblock our translators.

I've also updated the translation sources using this new workflow.

Finally, this changes to using an explicit Python3-based virtual environment rather than the system default.

Please merge this using a merge commit so Weblate doesn't explode.


Flohack74:
Hmm we shuld have talked about this before... Basically Weblate can use the pot files to update, or rather it is done by a Git hook on every pull. Only thing that would matter is that po and pot files are in the same directory. Then before Git leaves execution it will execute the post-update hook, running msgmerge and then the po files should be updated.

UniversalSuperBox:
Florian enabled translation updates from Weblate, this is no longer required.

PR #190, opened 2019-03-01
Always clean up

UniversalSuperBox:
This commit changes the Jenkinsfile so the post-build cleanup will always run no matter what.


UniversalSuperBox:
I found a better way to do this, back in a few moments

PR #194, opened 2019-03-01
Document enabling of /var/log/syslog

doniks:


UniversalSuperBox:
Writing to syslog is *not* normally disabled, but it is a step that people commonly need to perform to get their port off the ground. Is JBB's script still disabling rsyslogd by default?

doniks:
> Writing to syslog is _not_ normally disabled, 

ah, I was thinking that it's normally disabled in _UT_ ... but that is also total conjecture, I thought maybe it's to not wear flash, but I'm just guessing - I don't know whether that is normally so ... 

> Is JBB's script still disabling rsyslogd by default?

I'm not aware that it does or ever did. greping the script for "syslog" gives no hits.

Either way. How about just dropping the "Normally"? ie "The writing to the syslog is disabled. During porting it can be useful to enable this::"

UniversalSuperBox:
It seems really strange to me that it would be disabled at all... are you experiencing this problem with a current port?

doniks:
yes relatively recent. ut image 617 from ci (non-edge) and installed with jbbs script. to be clear with "disabled" I mean that the file /var/log/syslog simply did not exist. after the described steps, it does log. 

so, is the expectation that UT _should_ syslog? no flash wearing concern? In that case I'd withdraw the pr and we try to fix the rootfs

UniversalSuperBox:
Yes, unless `rsyslogd` is specifically disabled by placing `manual` into the `/etc/init/rsyslogd.override` file, logging should happen by default.

Marius has told me to do something like this in the past if syslog just won't run, but in general it should not be required.

doniks:
Well, no, it doesn't work without intervention. I can reproduce the original problem by
```
sudo rm /var/log/syslog
sudo reboot -f
```
then I can see that it doesn't work
```
ls /var/log/syslog
ls: cannot access '/var/log/syslog': No such file or directory
ps aux | grep syslog
phablet   2849  0.0  0.3 245332  5988 ?        S<l  21:51   0:00 pulseaudio --start --log-target=syslog
syslog   14080  0.0  0.0  31200   984 ?        Ssl  22:01   0:00 rsyslogd
root     14202  0.0  0.0   5360   804 pts/15   S+   22:01   0:00 grep --color=auto syslog
ls -ld /var/log/
drwxrwxr-x 10 root crontab 4096 mrt  1 21:50 /var/log/
initctl stop rsyslog
rsyslog stop/waiting
rsyslogd -d 
....
strm 0xb5301650: open error 13, file '/var/log/syslog': Permission denied
```
[full log](http://paste.ubuntu.com/p/J8P744H6w2/)

but when I create the file such that user `syslog` can write to it, then it's fine, but it can't create it by itself

PR #198, opened 2019-03-03
add note about usb raw to install instructions

Danfro:
Those instructions come from the note pinned to the welcome group.

### DIFF
--- a/userguide/install.rst
+++ b/userguide/install.rst
@@ -17,7 +17,7 @@ Start the installer and follow the on-screen instructions that will walk you thr
 
 .. note::
 
-    When using the snap installer sometimes device are not detected automaticaly. Then unplug your device, open a terminal and enter the following command:
+    If your device is not detected automatically when using the snap installer, then unplug your device, open a terminal and enter the following command:
 
     .. code-block:: text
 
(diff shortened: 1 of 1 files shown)


userguide/install.rst
```diff
@@ -15,6 +15,17 @@ Go to `the download page <https://github.com/ubports/ubports-installer/releases/
 
 Start the installer and follow the on-screen instructions that will walk you through the installation process. That's it! Have fun exploring Ubuntu Touch!
 
+.. note::
+
+    When using the snap installer sometimes device are not detected automaticaly. Then unplug your device, open a terminal and enter the following command:
```
UniversalSuperBox: It seems like we're switching between singular and plural here. Either stick with "devices" or reword the sentence to say "your device" in both cases.

"If your device is not detected automatically when using the snap installer, then ..."

Danfro: I thought of a general statement (devices not detected) and a instruction how to proceed (Then unplug...). But, yes your suggestion sounds good. Let's do it!

NeoTheThird:
Please revert this, it is no longer required and can lead to other issues. I removed that section for a reason!

UniversalSuperBox:
This was removed with the merge of #202

Danfro:
@UniversalSuperBox  Thanks for cleaning up. 

And thanks to everyone for sorting this out with the installer. The pinned message from the welcome group is gone too (thats why I put that note there in the first place). ;-)

PR #205, opened 2019-03-27
click: Update policy and manifest samples

fredldotme:
policy_version 1.3 is not correct as of now, use the distribution version instead.
Additionally add a second hook to the manifest example.


fredldotme:
On second thought I'm not sure if having a second hook in there makes sense for simple applications..

UniversalSuperBox:
I don't think it does... especially since "`foobar`" isn't very clear to its actual purpose.

UniversalSuperBox:
With #206 merged, you'll need to update your local branch (`git pull --rebase [remote] master`) and push again. Since you'll be rebasing, you'll need to force-push.

PR #213, opened 2019-04-28
Added script to generate page for Suru icons

bhdouglass:
Created a script to generate a page for Suru Icons, also includes all the icons and the generated page. I opted not to include the ubuntu-themes repo as a submodule in case the screws with the build on read the docs. If that's not desirable I can change it.


UniversalSuperBox:
It appears that submodules are supported as part of the new configuration format. Good time to migrate? https://docs.readthedocs.io/en/stable/config-file/v2.html

bhdouglass:
Well, it wouldn't be necessary to generate the icons page with every docs update, so that's probably overkill for right now. Unless of course we change our icon set often.

UniversalSuperBox:
How long does the generation take? As a rule I don't like generated files saved in version control.

bhdouglass:
The generation doesn't take long, so you could probably add it to the build. And I agree with that rule, but it probably won't change enough to warrant big changes to the build process. But of course, I'll leave that up to you.

UniversalSuperBox:
We absolutely need this, no doubt about that. I think that integrating this way will cause a couple problems though:

* Any constant text gets tossed into the translations... so we'll get translated icon names
* That's a lot of clone time we're adding...
* You're right, it's a waste of time to rebuild this on any change

Maybe the icons should be moved into another project ([Subprojects](https://docs.readthedocs.io/en/stable/subprojects.html)), that way we can rebuild them independently. Would only need a new repository, either of us can import it into Read the Docs and add it as a subproject.

bhdouglass:
With the subproject do you get the navigation the same between them? If not it might make more sense to just make an html page.

UniversalSuperBox:
Search is shared, we might have to do something a bit strange for navigation (the subproject goes at `docs.ubports.com/projects/icons/en/latest`, for example). Not a big deal.

bhdouglass:
Ok, closing this PR and starting https://github.com/ubports/icon-docs

PR #218, opened 2019-05-09
Add link to the kirigami documentation

JBBgameich:
Kirigami is included in the Ubuntu Touch image and should be documented :)


bhdouglass:
I think we should have a disclaimer that this is available but not the "Ubuntu Touch way" or something.

bhdouglass:
@UniversalSuperBox do translations need updating or anything before merging this?

PR #220, opened 2019-05-12
update translations.rst

Danfro:
Updated information about how strings are translated (technical information as background) and on how to translate.

Please have a closer look especially at the parts regarding the forum if they are still valid. I have got no experience with this way of translations. Does this really wants to be a third way of translation? That was what the old version suggested (how I understood it anyway).

Or should the forum be given as general information only on how to get help or discuss translation?

### DIFF
--- a/contribute/translations.rst
+++ b/contribute/translations.rst
@@ -20,11 +20,11 @@ Methods of Translation
 
 There are three common ways used to translate an app:
 
-* Weblate: A web based translation tool. UBports apps using Weblate can be translated `here <https://translate.ubports.com/>`_. There is an automatic transfer of new strings from an apps repo to weblate. Translated strings get back as commits to the repo from weblate.
+* Weblate: A web based translation tool. UBports apps using Weblate can be translated `here <https://translate.ubports.com/projects/>`_. There is an automatic transfer of new strings from an apps repo to weblate. Translated strings get back as commits to the repo from weblate.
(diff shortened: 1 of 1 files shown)


Danfro:
Ah, another point to review:

Do we really want to leave information here that translators can add their language by themselves on weblate? If yes we should include instructions. If no we should take that sentence out.

Flohack74:
@Danfro you have to re-phrase the text a bit. We cannot accept people translating directly in the po files. It will conflict often with Weblate, and there is no automatic conflict resoultion tool. Since we do not speak all languages, and we want to have the extended history and audit trail in Weblate, to better follow wrong translations, people must not do any changes in the po files (pot is OK). Instead, they should use the more comfortable features of Weblate.
Also, in the future there will be language teams that are allowed to commit translations. All outsiders only can suggest translations, but they are not automatically committed without review.

Danfro:
Hm, I 'thought' I left a note somewhere in there that .po translation does only apply to apps that are not translated via weblate. But, well, I can't find it. So obviously I forgot that.  :smile:
Sorry, of course that needs to be said explicitly. I will go over the text and add that information...

Flohack74:
Can you add a link to https://translate.ubports.com/projects/ where you put the note, explaining that people can find the actual Weblate projects at this link?

Danfro:
I do have UBports weblate linked twice already. Does it really need a third link?
But if you want to make sure people get there, of course I can do that. :+1: 

The existing links lead to [https://translate.ubports.com/](https://translate.ubports.com/) rather than to the projects page. The resulting page is pretty much the same though.

Danfro:
Ok, all links now point to the same page and I did add another link. I hope this is that what you had in mind...

PR #228, opened 2019-06-23
Update from Weblate

ubports-weblate:


UniversalSuperBox:
@Flohack74, I don't think this is needed.

PR #232, opened 2019-07-06
Add a page about the Calendars stack

mardy:
This is mostly some debugging instructions I found when debugging the
calendar synchronisation. The page should be expanded with other info,
but this is a start.

### DIFF
--- a/systemdev/calendars.rst
+++ b/systemdev/calendars.rst
@@ -1,0 +1,50 @@
+Working on the Calendar feature
+===============================
+
+The Calendar feature is provided by several components:
+
+* The `Calendar App <https://gitlab.com/ubports/apps/calendar-app>`_, providing the user interface
+* `Evolution Data Server <https://github.com/ubports/evolution-data-server-packaging>`_ (often shortened as "EDS"), which is the backend where your calendars are stored
+* `sync-monitor <https://github.com/ubports/sync-monitor>`_, the service responsible for managing the synchronisation with remote calendars
+* `SyncEvolution <https://github.com/ubports/syncevolution-packaging>`_, the service responsible for performing the synchronisation to a WebDAV/CalDAV remote calendar
+
+
+Debugging
+---------
+
+Calendar synchronisation issues
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+On the device, it's possible to run ``syncevolution`` in debug mode. To do so, open a terminal (it's easier if done from a remote shell -- you can gain a shell on the device using :doc:`/userguide/advanceduse/adb` or :doc:`/userguide/advanceduse/ssh`) and kill any existing processes:
+
+.. code-block:: bash
+
+        pkill sync-evo
+
+Then, start the ``syncevolution`` process with the proper environment variable:
+
+.. code-block:: bash
+
+        SYNCEVOLUTION_DEBUG=1 /usr/lib/arm-linux-gnueabihf/syncevolution/syncevo-dbus-server
+
+At this point, open the Calendar App (if it was not open already) and manually trigger a synchronisation (note that the Synchronisation action is only available if your device is connected to the internet): all the logs will appear in the terminal.
+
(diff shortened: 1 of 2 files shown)


systemdev/calendars.rst
```diff
@@ -0,0 +1,54 @@
...
+
+The Calendar feature is provided by several components:
+
+* The `Calendar App <https://gitlab.com/ubports/apps/calendar-app>`_, providing the user interface
+
+* `Evolution Data Server <https://github.com/ubports/evolution-data-server-packaging>`_ (often shortened as "EDS"), which is the backend where your calendars are stored
+
+* `sync-monitor <https://github.com/ubports/sync-monitor>`_, the service responsible for managing the synchronisation with remote calendars
+
+* `SyncEvolution <https://github.com/ubports/syncevolution-packaging>`_, the service responsible for performing the synchronisation to a WebDAV/CalDAV remote calendar
```
UniversalSuperBox: I wasn't aware that you could have blank lines between bullet points and keep them in a single list. Even so, this is the only file that does this. Would you mind removing the blank line to keep the style?

systemdev/calendars.rst
```diff
@@ -0,0 +1,54 @@
...
+
+
+Debugging
+---------
+
+Calendar synchronisation issues
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+On the device, it's possible to run ``syncevolution`` in debug mode. To do so, open a terminal (it's easier if done from a remote shell) and kill any existing processes:
```
UniversalSuperBox: This is a good place to link the pages "Shell access via ssh" and "Shell access via adb", like what's done in the other systemdev pages.

```
You can gain a shell on the device using :doc:`/userguide/advanceduse/adb` or :doc:`/userguide/advanceduse/ssh`. 
```

systemdev/calendars.rst
```diff
@@ -0,0 +1,54 @@
...
+Then, start the ``syncevolution`` process with the proper environment variable:
+
+.. code-block:: bash
+
+        SYNCEVOLUTION_DEBUG=1 /usr/lib/arm-linux-gnueabihf/syncevolution/syncevo-dbus-server
+
+At this point, open the Calendar App (if it was not open already) and manually trigger a synchronisation (note that the Synchronisation action is only available if your device is connected to the internet): all the logs will appear in the terminal.
+
+In some cases, the output from ``syncevolution`` might not be enough: for example, the raw HTTP data is usually not printed. Should you need to see that as well, then you'll have to modify a configuration file as well. ``syncevolution``'s configuration files are located under ``~/.config/syncevolution/``,  in a subdirectory whose name takes the form ``<provider-name>-<account-id>``.
+There might be stale directories as well, referring to old accounts which have been since deleted; to find out what accounts are still valid, you can invoke the ``account-console`` tool like this:
```
UniversalSuperBox: A single newline won't cause a paragraph break, if that is what you were trying to achieve here. If not, please don't break lines arbitrarily.

The semicolon is a little awkward. Maybe it would be better to end the sentence instead.

systemdev/calendars.rst
```diff
@@ -0,0 +1,54 @@
...
+
+        # level of detail for log messages:
+        # - 0 (or unset) = INFO messages without log file, DEBUG with log file
+        # - 1 = only ERROR messages
+        # - 2 = also INFO messages
+        # - 3 = also DEBUG messages
+        # > 3 = increasing amounts of debug messages for developers
+        loglevel = 11
+
+Note that in order for these changes to take effect, you'll need to restart the ``syncevolution`` process again, as explain above.
```
UniversalSuperBox: "As explained above" is a little awkward in this context. It might be good to add more sections to the page so you can link directly to the section (for example, `you'll need to //Restart the syncevolution process in debug mode//`, with "Restart the syncevolution process in debug mode" being your section title.

mardy: I don't think we should add another section until we also have more content. I'll fix the text to be "explained" and not "explain". Let me know if you can think of some improvements without creating more sections.

systemdev/calendars.rst
mardy: done

PR #248, opened 2019-08-28
Include the command to actually run graphical containered applications

tallero:
I had to get this information from a youtube video:
https://www.youtube.com/watch?v=aouooDea1D0

### DIFF
--- a/userguide/dailyuse/libertine.rst
+++ b/userguide/dailyuse/libertine.rst
@@ -59,9 +59,10 @@ Remove a package::
 Run desktop application
 -----------------------
 
-To run a graphical application you installed whose package installed a desktop file, i.e. ``gedit``, run in a terminal::
+To start a graphical application from the command line, for example ``gedit``, run in a terminal::
 
-  ubuntu-app-launch <CONTAINER_ID>_<desktop_file_name>_0.0
+  # ubuntu-app-launch <CONTAINER_ID>_<desktop_file_name>_0.0
+  ubuntu-app-launch xenial_gedit_0.0
 
 Files
 -----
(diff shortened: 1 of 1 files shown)


userguide/dailyuse/libertine.rst
```diff
@@ -56,6 +56,14 @@ Remove a package::
 .. note::
     If you have more than one container, then you can use the option ``-i CONTAINER-IDENTIFIER`` to specify for which container you want to perform an operation.
 
+Run desktop application
+-----------------------
+
+To run a graphical application you installed whose package installed a desktop file, i.e. ``gedit``, run in a terminal::
```
doniks: Maybe it's better to express that this is the way to start *from the command line*, which is not the normal way most users would do it. Also, maybe sentence could be split to be easier to read. How about 

To start a graphical application from the command line you can use ubuntu-app-launch. For example to start gedit, .....

userguide/dailyuse/libertine.rst
```diff
@@ -56,6 +56,14 @@ Remove a package::
...
+
+  ubuntu-app-launch xenial_gedit_0.0
+
+If you are in doubt about what the argument must be, just start typing ``xenial`` and then find all the possible ones with the ``tab`` key.
```
doniks: I think ``xenial`` here is the CONTAINERID, but I don't have a working device now to confirm. Can you check whether you get autocompletion also before you type any part of the appid? So, just ``ubuntu-app-launch[SPACE][TAB][TAB]". If so, please reformulate without "xenial", since other people might have different or multiple containers with other ids

tallero: Check now https://github.com/ubports/docs.ubports.com/pull/248/commits/c520955cde9eb58b4532ae0f17e112f4b261b562.

PR #256, opened 2019-09-28
Update libertine.rst

PJK2011:
Add graphical interface instructions.


doniks:
Thanks @PJK2011 I'm happy to merge as is. If you want to do another round, maybe it would be an easier read for most users to just have the simple to read/simple to do/ good enough for 99% of use cases section from the UI on top and put all the nerdy command line stuff at the bottom instead of interlacing normal UI and command line

PR #258, opened 2019-10-04
Warn that boot flashing requires a backup

tallero:
I am adding this because doing it on the `dev` channel bricked my hammerhead.

### DIFF
--- a/userguide/dailyuse/anbox.rst
+++ b/userguide/dailyuse/anbox.rst
@@ -26,7 +26,7 @@ How to install
 .. warning::
     Installing Anbox is only recommended for experienced users.
 
-- Make sure your supported device runs on 16.04 (Anbox doesn't work on 15.04); be sure to have a `_backup <../backup.rst>`_ of the device.
+- Make sure your supported device runs on 16.04 (Anbox doesn't work on 15.04); be sure to have a `_backup <https://askubuntu.com/questions/602850/how-do-i-backup-my-ubuntu-phone>`_ of the device.
 - Open a terminal on your host and set some device specific variables by running ``export CODENAME="turbo" && export PARTITIONNAME="bootimg"``, but replace the part between the quotes respectively with the codename and name of the boot partition for your device. See the above list.
 - Activate developer mode on your device.
 - Connect the device to your host and run the following commands from your host (same terminal you ran the ``export`` command in)::
--- a/userguide/dailyuse/anbox.rst
+++ b/userguide/dailyuse/anbox.rst
@@ -26,7 +26,7 @@ How to install
 .. warning::
     Installing Anbox is only recommended for experienced users.
 
-- Make sure your supported device runs on 16.04 (Anbox doesn't work on 15.04); on `dev` channel be sure to have a backup of the device.
+- Make sure your supported device runs on 16.04 (Anbox doesn't work on 15.04); be sure to have a `_backup <../backup.rst>`_ of the device.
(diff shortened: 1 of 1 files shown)


tallero:
>could it be that you were just unlucky? dev isn't meant for end users and it can break, nor is anbox (as of now) for the faint of heart. however once one heeds those warnings, is there really a general rule that one should not use dev+anbox? if so, I think we would need more evidence other than "it has happened once"

you use anbox on dev usually?
I thought anbox `img`s had binary compatibility on the stable `system.img`, I did not think of a bug, really.

doniks:
I have never used anbox. There is no 'binary' difference between stable or dev images. The difference is purely timing and amount of testing. Given testing results todays dev image will be called rc tomorrow and stable some days later.

I assume the difference of the anbox image is some kernel params or maybe some changes to initrd.

userguide/dailyuse/anbox.rst
```diff
@@ -26,7 +26,7 @@ How to install
 .. warning::
     Installing Anbox is only recommended for experienced users.
 
-- Make sure your supported device runs on 16.04 (Anbox doesn't work on 15.04)
+- Make sure your supported device runs on 16.04 (Anbox doesn't work on 15.04); on `dev` channel be sure to have a backup of the device.
```
doniks: I'm still not convinced that there is anything specific about the channel. If we want to warn about backups, I think we should do it without the dev disclaimer. 

Btw, note that there is a recently merged warning about re-flashing anbox boot img

tallero:
>I assume the difference of the anbox image is some kernel params or maybe some changes to initrd.
I believed `dev` and `stable` ran (possibly) two different kernel builds, so I expected the initramfs in the image being tailored to the `stable` kernel.

userguide/dailyuse/anbox.rst
tallero: >Btw, note that there is a recently merged warning about re-flashing anbox boot img
what do you mean?

userguide/dailyuse/anbox.rst
```diff
@@ -26,7 +26,7 @@ How to install
 .. warning::
     Installing Anbox is only recommended for experienced users.
 
-- Make sure your supported device runs on 16.04 (Anbox doesn't work on 15.04)
+- Make sure your supported device runs on 16.04 (Anbox doesn't work on 15.04); be sure to have a `_backup <../backup.rst>`_ of the device.
```
doniks: The link doesn't seem to go anywhere. Do we have a backup doc? 

https://ci.ubports.com/job/docs.ubports.com/job/PR-258/3/artifact/_build/html/userguide/dailyuse/anbox.html

Also, there is some extra underscore.

Lastly. maybe it could be it's own bullet point?

userguide/dailyuse/anbox.rst
doniks: https://github.com/ubports/docs.ubports.com/pull/262

tallero:
> could it be that you were just unlucky? dev isn't meant for end users and it can break, nor is anbox (as of now) for the faint of heart. however once one heeds those warnings, is there really a general rule that one should not use dev+anbox? if so, I think we would need more evidence other than "it has happened once"

Apparently this issue could be the first reporting of this series of issues with OTA-11 and anbox: 
[1264](https://github.com/ubports/ubuntu-touch/issues/1264)
[1286](https://github.com/ubports/ubuntu-touch/issues/1286)
[1280](https://github.com/ubports/ubuntu-touch/issues/1280)

PR #263, opened 2019-10-28
Add information about the possibility to switch between channels inside the OS

Flaburgan:


Flaburgan:
I just saw that `edge` is now available there to (at least with my phone on the RC channel). Do we want to mention it in the docs, or is this a very unstable channel and we don't want to promote it too much?

PR #265, opened 2019-10-30
Add indications about adb on the computer and adb on the phone

Flaburgan:


userguide/dailyuse/anbox.rst
```diff
@@ -54,12 +54,22 @@ How to install
 
 - Done! You might have to refresh the apps scope (pull down from the center of the screen and release) for the new Android apps to show up.
 
+.. note::
+    You now have an adb server running inside your phone. This guide asks you to run some ``adb`` commands. Carefully check on which device you have to run them. If you want to be sure to not mess up between your phone and your computer, connect to your phone using ``ssh`` instead of ``adb shell`` and keep ``àdb`` for commands from your phone only. See :doc:`this guide </userguide/advanceduse/ssh>` about connecting to your phone using ``ssh``.
```
doniks: there's some accented character on one of the a's : àdb

doniks:
I really like the explanations about having a local adb server and how to check it. 

But I am a bit hesitant about the ``ssh`` part. It seems it makes the docs just harder to read and more complicated. Are you sure we need this?

Flaburgan:
Hm, it's a tip which helped me. What made me not sure about it was the fact that the `ssh` tuto is in the "Advanced use" category while Anbox is in the "Daily use". But honestly, I would have moved Anbox in the advanced use, I'm a developer myself and still struggling with it :stuck_out_tongue: 

What do you think about another Note section at the very beginning, saying "You will need to connect to your phone to run commands locally in this guide. You can do that by running `adb shell` or by [setting up an ssh connection to your phone] (link to the doc)" and then remove the "tip" part in the second note, but keeping the `# or ssh` in the instructions?

doniks:
> Hm, it's a tip which helped me. 

> What made me not sure about it was the fact that the `ssh` tuto is in the "Advanced use" category while Anbox is in the "Daily use". But honestly, I would have moved Anbox in the advanced use, I'm a developer myself and still struggling with it stuck_out_tongue

Agree, anbox should move to advanced. Intuitively, I would draw the line at: "Do you need to break out the command line?" Yes -> put it into advanced. 

(Even the Libertine topic could be argued about. I think it just barely makes the cut for "daily" now that it finally has some non-command-line explanations)

> What do you think about another Note section at the very beginning, saying "You will need to connect to your phone to run commands locally in this guide. You can do that by running `adb shell` or by [setting up an ssh connection to your phone] (link to the doc)" and then remove the "tip" part in the second note, but keeping the `# or ssh` in the instructions?

Honestly, the `# or ssh` in the instructions is the part I like least. How about 

```
adb shell # connect from your host computer to your UT device
foo
bar
adb shell # connect from your UT device to the Android container
```

Then the tips on top can refer to "for connecting from your host computer to the UT device you can also use ssh ...." but there is more "loose coupling". One can easily skip over all ssh stuff and only read it without and one can also find where to plug in the ssh if needed. 

PS: I'm hesitating to mention it, because I am obviously trying to not stick too many details in here ...  but there is also `adb -d DEVICEID` ;)

Btw. Are you on telegram? Feel free to ping me: I'm dohniks there. I can also add you to the UT docs group if you want.

PR #267, opened 2019-11-30
Add paragraph about android storage

tallero:
Add paragraph about android storage.
Add example storage access configuration


tallero:
Yes I had the same problem but I couldn't understand why was displayed that way, since the syntax was the same as other code blocks. With the latter it was not working, too.

Will try with the double colon.

PR #271, opened 2019-12-17
Update userguide.po

kalinovka:
Hello! I made some translations into Russian.


Flohack74:
Please use translate.ubports.com, we cannot accept direct translations of po files, that does not fit in our workflow.

PR #272, opened 2020-01-01
Add language switching option: zh_CN

Flowertome:
Although translation files are already available,
 there is no clickable language switching option in docs.ubports.com.


Flowertome:
@doniks Thank you

doniks:
Thank you!

PR #280, opened 2020-01-12
Reminder to update app version after policy update

nboumalham:
It seems like this is one of the most asked question in forums. People update their policy and don't update the app version which doesn't then take into account this change.


UniversalSuperBox:
Hi @nboumalham, are you able to update this PR to address the requested changes?

PR #286, opened 2020-02-07
Edit to User Metrics guide

userj:
Some updates to the user metrics guide to make it more accurate according to another blog that I 
found. 

* There are actually two methods for modifying metrics
* Fixed error regarding the arguments passed to `update` method
* Mention the deletion of metrics
* FluffyChat as an example of keeping counts longer than one day


userj:
Oops. Ok I think it's the updated with the current master now.

PR #289, opened 2020-02-18
Origin/patch 1

userj:
It seems that the editor used before changed all the double quotations somehow in the previous merge attempt.

* Updated accuracy of the methods available to metrics
* Removed some unnecessary whitespace


userj:
Sorry, something else needs updating.

PR #300, opened 2020-04-08
Adapt docs for Halium-enabled devices

Flohack74:
This will indicate that reflashing the Anbox kernel is only needed for certain devices from the list. I don´t know however if that will be clear enough. Merge this after OTA-12 and additionally I suggest to remove the kernel downloads from cdimage so we dont lead users into unnecessary flashing.


doniks:
Lgtm. Why does this have to wait for ota 12. Both seem universally true to me

PR #308, opened 2020-05-23
python 3

doniks:


doniks:
on Ubuntu 20.04 the current ./build.sh doesn't work. I needed these changes.

I think 20.04 has deprecated python2. Someone with an older Linux and python 2 please test this!

doniks:
wait, that doesn't sound right

PR #311, opened 2020-05-25
Update install.rst

PigCode:
Improve the installation process

### DIFF
--- a/userguide/install.rst
+++ b/userguide/install.rst
@@ -150,7 +150,7 @@ Meizu devices are pretty much stuck on Flyme. While the MX4 can be flashed succe
 8. Turn your device completely off, but do not connect it yet.
 9. Press the button labeled "Download".
 10. Perform a final sanity-check that you selected the "Firmware Upgrade" option, not "Download Only".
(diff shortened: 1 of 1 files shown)


mymike00:
I don't think pressing the down volume is always needed. I'd rather add it as an option if just plugging doesn't work.

PigCode:
> I don't think pressing the down volume is always needed. I'd rather add it as an option if just plugging doesn't work.

Meizu MX4 Android Edition must pressing the down volume

mymike00:
something like "plug the usb cable. if nothing happens, disconnect the cable and reconnect it while keeping the volume down button pressed: some phones require that, some others don't"

Better solution is to test all the devices and find out which ones require the button pressed and which ones don't, but I'd take some more time...

doniks:
Maybe like: On some devices (such as x) you need to do this. On other devices (such as y) do that.

Now a user with device z comes around, and should: we don't know for sure, but trying this or that are probably good suggestions

doniks:
ping @PigCode

doniks:
Seems the contributor has disappeared. I'm closing this for now. @PigCode  if you still/again want to fix this up, feel free to reopen!

PR #317, opened 2020-06-12
Add adb shell commands

gfda:
Adding some useful adb shell commands to adb.rst


peter-pycom:
It looks a bit lost, just stuck in the middle. Mhm, actually the paragraphs above also already look like random remarks cobbled together. Oh well, lets add this now and make it pretty another day

PR #319, opened 2020-06-30
Explain CI artefact

doniks:
plus two small adjustments


NTpspE:
@doniks Aha, thanks for adding the CI information, I was able to see your version by following the instructions and that makes a lot of sense! :)

PR #322, opened 2020-07-08
Update writeable-dirs.rst: Qt app name and clickable app name need to match

mlavik1:
Adding a note about the Qt application name needing to match the clickable application name for these paths to work.
In my case they were not the same, and trying to write to QStandardPaths::AppDataLocation, since the directory did not exist and was not a writable directory.


doniks:
thx!

PR #336, opened 2020-08-31
Some minor edits and integration of new page to incorporate points fr…

abkro:
Please check that the info herein is correct.

### DIFF
--- a/porting/device-configuration.rst
+++ b/porting/device-configuration.rst
@@ -1,0 +1,78 @@
+Device configuration
+--------------------
+
+The rootfs (root file system) you have installed on your device contains a number of configuration files with some standard settings that do not necessarily conform to your device. These must be modified to fit your device, but you are not allowed to do so by mounting the rootfs read/write and editing them directly on the device. Instead, working on your host pc, you edit the files there and inject them into your system at build time as outlined below.
+
+Note that as a rule of thumb, the method described below applies if the file you need to edit can be found in the /etc directory (or a subdirectory of this) on your device. You should not attempt to overwrite files located elsewhere with overlay files using this method.
+
+In your device directory, create a subdirectory named 'ubuntu'. Collect the files you wish to inject into your build in this directory. Relevant files are for example (but this list is incomplete):
+    * 70-android.rules (the udev rules for your device)
+    * android.conf (for display scaling)
+    * touch.pa (for pulseaudio sound configuration/initialization)
+
+These files are then injected by adding a code block to the file 'device.mk' in your device directory. For the three files above add the following code::
+
+    ### Ubuntu Touch ###
+    PRODUCT_COPY_FILES += \
+        $(LOCAL_PATH)/ubuntu/70-android.rules:system/halium/lib/udev/rules.d/70-android.rules \
+        $(LOCAL_PATH)/ubuntu/android.conf:system/halium/etc/ubuntu-touch-session.d/android.conf \
+        $(LOCAL_PATH)/ubuntu/touch.pa:system/halium/etc/pulse/touch.pa 
+    ### End Ubuntu Touch ###
+
+Explanation:
+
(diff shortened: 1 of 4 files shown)


porting/device-configuration.rst
```diff
@@ -0,0 +1,78 @@
+Device configuration
+--------------------
+
+The rootfs (root file system) you have installed on your device contains a number of configuration files with some standard settings that do not necessarily conform to your device. These must be modified to fit your device, but you are not allowed to do so by mounting the rootfs read/write and editing them directly on the device. Instead, working on your host pc, you edit the files there and inject them into your system at build time as outlined below.
```
doniks, line 4: "you are not allowed" sounds a bit arbitrary. there is a reason for it. How about something like

 The rootfs (root file system) you have installed on your device contains a number of configuration files with some standard settings that do not necessarily conform to your device. These must be modified to fit your device. The rootfs itself is read only and identical for all devices. Below it is described how to adjust the configuration with overlay files.

porting/device-configuration.rst
```diff
@@ -0,0 +1,78 @@
...
+
+    The target paths for the files mentioned above are *not* randomly chosen. You must use the specified paths. 
+
+.. note::
+
+    When you specify target path 'system/halium/etc/myfilename' your file 'myfilename' will end up in the '/etc' directory of your device (i.e. without the leading 'system/halium')
+
+When you have made the adjustments you need and prepared your source as described above, you have to rebuild your system.img: ``mka systemimage``. When rebuilding the system image after small changes like these, you need not ``mka clean`` first. However, changes to PRODUCT_PROPERTY_OVERRIDES might not get detected by the build system. Go to your output folder, enter the system folder and delete build.prop in order to get it regenerated.
+
+----------------
```
doniks, line 42: this looks like the syntax to create a section headline ... is this here intentional? If you wanted a headline here, it's missing. Otherwise I suggest to remove the dashes.

porting/device-configuration.rst
```diff
@@ -0,0 +1,78 @@
...
+At the end of the file, append this::
+
+    ### Automatically load the audioflinger glue
+    .ifexists module-droid-glue-24.so
+    load-module module-droid-glue-24
+    .endif
+
+Now complete steps 3 and 4.
+
+---------------------
```
doniks, line 76: again

porting/device-configuration.rst
```diff
@@ -0,0 +1,78 @@
...
+
+TODO: Add more from the porting faq and porting check list.
```
doniks, line 78: please add the links

porting/installing-16-04.rst
```diff
@@ -48,14 +48,14 @@ The password will be the one that you set while running halium-install.
 Common Problems
 ---------------
 
-If you have any errors while performing these steps, check see if any of the following suggestions match what you are seeing. If you have installed successfully, skip down to `Continue on`_.
+If you have any errors while performing these steps, check see if any of the following suggestions match what you are seeing. If you have installed successfully, skip down to `The Next Steps`_.
```
doniks, line 51: check to see

doniks (changes requested):
Thanks a lot @abkro  for this! Really very much appreciated. 

Some very minor comments below.

porting/device-configuration.rst
abkro: Resolved, but in new branch. New PR will be submitted.

porting/device-configuration.rst
abkro: Ditto

porting/device-configuration.rst
abkro: Ditto

porting/device-configuration.rst
abkro: Ditto

porting/installing-16-04.rst
abkro: Unclear. Don't understand what is required here?

abkro: Yes! Strange. I thought I had corrected that. It was in the original version. Fixing it now...

PR #349, opened 2020-10-08
Add redirects from old porting pages with Rediraffe

UniversalSuperBox:
The merge of #342[1] brought some awesome new content, but it also removed some pages. We have links to these pages elsewhere (other documentation, our website, blogs) and it would be a shame to send those to a 404.

To fix this, set up Rediraffe to check for files removed in the current branch. This will warn us in CI output when a file is removed without a redirect. By adding the 'rediraffecheckdiff' build at the end of build.sh, people can easily check their own docs.

Rediraffe also allows *creating* redirects from the newly nonexistent pages, so we'll start with the recently-removed porting pages in redirects.txt.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,6 +3,7 @@ def install_dependencies = {
 }
 def build_docs = {
     sh 'python3 -m sphinx -Wab html . _build/html/'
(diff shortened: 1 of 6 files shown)


NeoTheThird (approved):
good idea

doniks (approved):

PR #352, opened 2020-10-14
Switch to Python3 in build script

jonnius:
This MR
* switches build script from Python 2  to Python 3
* Makes the build script fail properly on errors instead of just continue with the next step
* Prints the path to the entry point to open for navigating

Python 2 reached end of life and is not preinstalled on most distros nowadays. Python 3 in contrast should be available and installed everywhere.


doniks (approved):

doniks:
thanks Jonny!

PR #353, opened 2020-10-14
Avoid expanding export PATH command in crossbuilder setup

jonnius:
```
echo "export PATH=$HOME/crossbuilder:$PATH" >> ~/.bashrc
```
results in `$PATH` being replace by the current content of the `PATH` variable while it is appended to `~/.bashrc`. You definitely do  not what that as it will overwrite whatever is in `PATH` the time it is applied in the future. This MR fixes that by add `'`s arround the line to appended to `~/.bashrc`.


doniks (approved):
Yes!

PR #368, opened 2020-12-01
Add a hint about never running installer with sudo

NeoTheThird:
Many medium-experienced linux users have the knee-jerk reaction of running stuff with sudo that doesn't work on the first try. This is a **bad idea** and needs to be discouraged. The installer does not want to be run with sudo. Believe me, i wrote the damn ~~bill~~ installer.


doniks:
lol. okok, taking the advice straight from the horses mouth!

doniks (approved):

PR #373, opened 2021-01-04
make docs.ubports.com/porting a valid link

Fuseteam:


doniks:
as far as I can *test* it, it looks ok to me, but I'm wondering about two things:

1. the capital I in Index. Is this going to work that http://docs.ubports.com/en/latest/porting will properly get redirected to this page? I think it should get redirected, but I'm wondering whether it will work with the capital I. In all other places in the docs, it's lower caps index.rst
2. and relatedly, if this redirecting would work properly, can't we then leave the word "Index" out completely? I mean, e.g., link to `/porting` instead of `/porting/Index` or `/porting/Index.rst` 

I can't find a way to properly test this, because in Jenkins, and when I build it locally, it simply shows me the directory listing, e.g., https://ci.ubports.com/job/docs.ubports.com/job/PR-373/1/artifact/_build/html/, but in the live website this url gets redirected to the index

Unless, I'm missing something, I would much prefer to see it renamed to lower caps index and then removed where it isn't needed.

Fuseteam:
@doniks alright i'll make those changes then

Fuseteam:
i noticed in systemdev/kernel-hal.rst /index is used for `appdev/index` and `systemdev/index`
should i still remove index from porting/index?

Fuseteam:
before i forget removing `/index` from redirections.txt redirects to `porting` even if it's specified as `porting/`

doniks:
> i noticed in /index.rst `/index` is used for `appdev/index` and `systemdev/index`
> should i still remove index from porting/index?

I see. Ok, let's keep it the way it is now



> before i forget removing `/index` from redirections.txt redirects to `porting` even if it's specified as `porting/`

ah. ok. good!

redirects.txt
```diff
@@ -1,4 +1,6 @@
-"porting/introduction.rst" "porting/Intro.rst"
+"porting/introduction.rst" "porting/index.rst"
+"porting/Intro.rst" "porting/index.rst"
+"porting/Index.rst" "porting/index.rst"
```
doniks, line 3: this one we don't need now, right? I mean before this PR, there was never a mention of "porting/Index.rst", so we don't need to redirect it.

doniks (changes requested):

redirects.txt
Fuseteam: without this it won't build for me locally

peter-pycom: strange. maybe after "make clean", like rm -rf _build or smt like that?

Fuseteam: hmmm with a squash rebase it does build fine without it, should i just force push?

peter-pycom: ok :+1:

Fuseteam: > strange. maybe after "make clean", like rm -rf _build or smt like that?

oh interesting it appears related to _build

doniks: Yeah, I've seen smt similar. I guess the build process doesn't delete files from the build folder

doniks (approved):

PR #377, opened 2021-01-31
Sshd start

doniks:


userguide/advanceduse/ssh.rst
```diff
@@ -43,20 +43,25 @@ If you are using a Linux-based phone (like the Pinephone)::
 
     sudo service ssh start
 
+To have the ssh server start after every boot on a Linux-based phone::
```
UniversalSuperBox, line 46: "Linux-based phone" seems unclear to me. Android is Linux-based, and so is Ubuntu Touch... everywhere.

UniversalSuperBox:
https://gitlab.com/ubports/core/hybris-support/lxc-android-config/-/merge_requests/42 fixes the problem that requires the remount and removal for real.

doniks:
let me acknowledge reality ... apparently I cant' find a way to express this properly

PR #396, opened 2021-03-12
Intro documentation language reworked

comradekingu:


about/introduction.rst
```diff
@@ -1,23 +1,23 @@
-Introduction
-============
+Intro
```
doniks, line 1: mhm ... the full in the header please

doniks (changes requested):

about/introduction.rst
comradekingu, line 1:
```suggestion
Introduction
```

doniks (approved):

Issue #388, opened 2021-03-07
multiple broken links on docs.ubports (bq/meizu)


Matte26:
There are currently 4 out of 5 links to the legacy devices rom that are not working, redirecting users to an error page.
In particular, Bq E4.5 (krillin), Bq E5 HD (vegetahd), Bq M10 HD (cooler), Bq M10 FHD (frieza) are not working.
Meizu MX4 (arale) is the only currently working link.

lduboeuf:
here ? : https://docs.ubports.com/en/latest/userguide/install.html#install-on-legacy-android-devices

Matte26:
Yeah, those are the links I was referring to, forgot to link the page, my bad

doniks:
> This is more or less uncharted territory.

it seems it became a bit more uncharted :( those were strange links to begin with ... some binary file on someones google file share ... 

fwiw, [ubuntuusers.de](https://wiki.ubuntuusers.de/Ubuntu_Touch/flash_tool/#Android-auf-ein-BQ-Geraet-flashen) is now linking to [some other forum](https://www.android-hilfe.de/forum/bq-allgemein.2373/software-repository-fuer-treiber-tools-firmware.964197.html) where there supposedly is a link to a telegram channels where some binary files are being shared ...

Matte26:
Checked it out, if I'm not mistaken, that telegram group only hosts the android roms, not the UT ones needed to make the first conversion of legacy devices before updating them with ut installer

Matte26:
As per ubuntuusers.de, the links they provide to those legacy devices ut fw, are the very same as the ones os docs.ubports, therefore broken as well

Fuseteam:
seems vegetahd is available in web archive: http://web.archive.org/web/20200820100403if_/https://storage.googleapis.com/otas/2014/Smartphones/Aquaris_E5_HD/Ubuntu/VEGETA01A-S11A_BQ_L100EN_2017_170207.zip
credits to @amyosx for finding it

TartanSpartan:
This is the cooler legacy image. The upload will only persist for a week, but hopefully that should suffice.

https://we.tl/t-uGsdbSqtFj

doniks:
Thanks guys! Images have been placed here: https://cdimage.ubports.com/stock-bq-images/

Does one of you want to make a PR to fix the docs? I could do it, but I usually struggle to find someone to review my PR's, so it will be fastest if someone else proposes and I review.

[...]

PR #399, opened 2021-03-12
Draft: Install documentation language reworked

comradekingu:


userguide/install.rst
```diff
@@ -146,18 +156,19 @@ Meizu devices are pretty much stuck on Flyme. While the MX4 can be flashed succe
 
 .. warning::
 
-    If you select DOWNLOAD ONLY rather than FIRMWARE UPGRADE, you will end up with a useless brick rather than a fancy Ubuntu Touch device. Be sure to select FIRMWARE UPGRADE.
+    You must select FIRMWARE UPGRADE, and not DOWNLOAD ONLY. Otherwise you brick your device.
 
-    Also, after selecting the ``*Android_scatter.txt``, the drop-down menu choice is reset to DOWNLOAD ONLY. Be sure to select it after selecting the file.
+    Also, after selecting the ``*Android_scatter.txt``, the drop-down menu choice is reset to DOWNLOAD ONLY.
+    Be sure to select it after selecting the file.
```
comradekingu, line 162:
```suggestion
    Make sure it is set to FIRMWARE UPGRADE after selecting the file.
```
?

doniks:
@comradekingu did you notice that this PR failed to build? Actually, all of your open PR's are failing right now. Two of them with a weird git error that I don't understand, but the others with genuine problems in your changes. Please check.

About this PR 399 specifically, It seems you are proposing some changes, but you are also opening a question on it yourself (I think also on another PR). I, personally cannot answer it myself. What should we do? If you want to still gather feedback and work on this, then maybe mark it as draft. Otherwise, maybe remove the particular passage that you are unsure about (for now) and we merge without ...

comradekingu:
@doniks I really don't want to guess which one of two options bricks user devices. From my point of view the distinction for which one to use isn't clear at all. That is the one thing to fix, if anything. Will mark as draft.

doniks:
I see. Maybe some ppl here https://github.com/ubports/docs.ubports.com/issues/388 have used this tool and can answer

doniks:
So .... what do we do? It seems that so far no answer has emerged. I agree that we shouldn't light heartedly publish statements that can brick peoples devices. 

My suggestion would be to simply leave the questionable sections unchanged. In that way we are not making matters worse on that note, but we would improve the other parts of the text with your contribution

doniks:
seems to me this is stalled, so I'm closing. feel free to reopen if you intend to proceed!

PR #419, opened 2021-04-04
apparmor: Add policy group description for nfc

fredldotme:
Short and to the point.


appdev/platform/apparmor.rst
```diff
@@ -205,6 +205,16 @@ Usage: common
 The networking policy group allows an app to contact network devices and use the `download manager`_.
 
 
+nfc
+----------
```
doniks, line 209: same length pls

doniks (changes requested):

doniks (approved):

PR #439, opened 2021-06-23
Improve dark theme and images

capsia37:
Hi @doniks ,
here we have some more improvements for the dark theme and svg compression.

- Yumi is now lighter in the dark theme
- Removed white backgrounds from pictures for dark theme
- Apply lossless compression to svg images for faster loading


doniks (approved):

doniks:
hahahaha. I only notice now that in dark mode it's a complete custom logo with a sleeping yumi. This is a cute Easter egg :D 

I think it could do with a little more color and contrast though. Right now it's pretty much dark grey on black and no orange at all.

doniks:
Oh, and in case you were to update the app dev guide images again, could you do me a favor and also rename them. 1.png 2.png 12.png etc isn't particularly expressive. It would make reviewing much easier if they had speaking names

doniks:
@capsia37 are you reading this here?

I just tried to build locally and I get

ERROR: furo 2021.3.20b30 has requirement sphinx~=3.0, but you'll have sphinx 4.0.2 which is incompatible.

when I run ./build.sh after I deleted my local ~/ubportsdocsenv/

with the old dir, I got 

Theme error:
no theme named 'furo' found (missing theme.conf?)

doniks:
Also, I notice now that the old theme had Previous and Next buttons at the bottom and the new one hasn't. Now, for some (many?) parts of the docs those were probably useless/awkward anyway, but for everything in the porting section, they were really helpful. Porters would (in my opinion) read the whole guide at least once front to back, or at least skip through them, because it really is the *sequence* of things you need to do when porting. 

Is there an easy way to get the buttons back? I think even if  we have the buttons everywhere, including places where they don't make too much sense it would be a better compromise, than not having them in the porting guide

capsia37:
Hello @doniks 

> ERROR: furo 2021.3.20b30 has requirement sphinx~=3.0, but you'll have sphinx 4.0.2 which is incompatible. 

Have you tried to reinstall the requirements for the docs to build
```
virtualenv ~/ubportsdocsenv
. ~/ubportsdocsenv/bin/activate
pip3 install -r requirements.txt
```

If the error is still occuring, you could also delete the sphinx and sphinx-rtd-theme packages using pip commands.

> Is there an easy way to get the buttons back? I think even if we have the buttons everywhere, including places where they don't make too much sense it would be a better compromise, than not having them in the porting guide

Yes, I can get them back, there was an option in the theme if I remember right.

For the Yumi dark logo, I think I can increase the contrast.

Thank you!

PR #452, opened 2021-08-28
Porting review

doniks:


doniks:
Hey @abkro this contains almost all my feedback from here https://github.com/ubports/docs.ubports.com/pull/441

It should all be just small stuff, except this one https://github.com/ubports/docs.ubports.com/pull/452/commits/e22979716fec7457ad267992b835e0926641c518 please check that one carefully.

porting/build_and_boot/index.rst
```diff
@@ -1,17 +1,9 @@
...
-
-For the remainder of this guide, information pertaining to Halium-9.0 is also intended to apply to newer versions of Halium. Note that as newer versions evolve, it is quite possible that certain differences will become apparent. As long as such differences remain few and relatively small, new sections will not be added to the guide, but as the number and magnitude differences grow, this option must eventually be chosen. 
-
-There is a separate section describing how to install the build. This section covers both Halium versions. Make sure to note the differences here also.
-
-Furthermore, a section describing an alternate method of building for Halium-9.0 based on Gitlab CI scripts has been included.
-
-The final section is concerned with booting and debugging the boot process.
+This depends on the Halium version and the porting method. Please refer only to the sections that are relevant to the method you are following.
```
abkro: Re-reading this now, I agree that it may be too long. Most of the time the porter who is familiar with the buide will go straight to the relevant section. These paragraphs were intended for a "first read" to give an outline of this chapter and how to use it. We can shorten it, but I worry that people may mess up when they get to the install section. What do you think? The rest is clearly labeled with the relevant halium version, but that section covers both. Is it sufficiently clear?

porting/build_and_boot/H7_setup_sources.rst
```diff
@@ -1,7 +1,7 @@
 Halium-7.1 - Setting up the sources
 ===================================
 
-The sources necessary for the build depend on the porting method you will be following. 
+The sources necessary for the **full system image build** for Halium-7.1.
```
abkro, line 4: Small adjustment, maybe: 
Halium-7.1 requires a **full system image build**.

porting/build_and_boot/H9_gitlab_build.rst
```diff
@@ -1,10 +1,12 @@
 
 .. _Gitlab-CI:
 
-Halium-9.0 - Using Gitlab CI scripts
-====================================
+Halium-9.0 - Standalone kernel method
```
abkro, line 4: OK. Here I want to keep the title because:
When discussing with Nikita, after starting to write about standalone kernel builds, he made a clear distinction between this and the Gitlab CI scripts build, as he called this. The difference is that with the former, you build a linux kernel (not halium-boot.img). You then need to manually follow a few steps to construct the halium-boot.img from the kernel. Then you use this together with the GSI in the same manner as you would if you had built the halium-boot.img in the standard way in the first place. After trying out the other method, which he recommended, I discovered that the chances of straying off the path and making mistakes was far smaller. So with this method, although you are just downloading the kernel source, you are actually building more than the standalone kernel.

porting/build_and_boot/H9_gitlab_build.rst
```diff
@@ -1,10 +1,12 @@
...
+=====================================
 
-For some devices, Halium-9.0 (and later) ports can be built based on just the kernel source code for your specific device and SoC (system on chip). This method makes use of scripts that download and prepare the kernel source, build the boot image and combine this together with other necessary components, thus creating the files needed which can subsequently be flashed to the device.
+For some devices, Halium-9.0 (and later) ports can be built based on just the standalone kernel source code.
```
abkro, line 7: This looks fine, provided you remove the word 'standalone', as explained above.

porting/build_and_boot/H9_gitlab_build.rst
doniks: mhm. well. I wonder whether that distinction is relevant enough. Since we're only describing one method for only kernel anyway, I would suggest to focus on the main aspect: just a kernel, rather than on particular technical aspect (it uses gitlab). I mean there are always more variants.(If I remember right Alfred used AOSP instead of Lineage). But ok. My main point is use a consistent name all through out. It took me a while to figure out that the third method mentioned at the start actually WAS the CI method. When I first read it I thought that the CI was an extra fourth one, but then I was lost because I couldn't find the third one anymore. Can you suggest a single name to be used everywhere?

porting/build_and_boot/H9_gitlab_build.rst
doniks: done

porting/build_and_boot/H7_setup_sources.rst
doniks: done

porting/build_and_boot/index.rst
doniks: I've reformulated it a little. pls recheck

porting/build_and_boot/H9_gitlab_build.rst
abkro: After re-reading this and thinking it through, I agree with your suggestions. The Gitlab-CI script vs. standalone kernel distinction is something the devs see, but which the average porter does not need to be concerned with, as long as we don't describe both anyway. 

However, I also noticed that in my PR this summer, I forgot to edit the introduction where the methods are briefly described. This is irrelevant now, but that part needs to be adjusted a little to bring it in line with the rest, if we are going to go with 'full system image build', 'halium-boot method' and 'standalone kernel method'.

The important thing to communicate, as I see it, is that Halium-9.0 can use all three methods, whereas Halium-7.1 can only use the first. And if the full system image build method is only described for halium-7.1 (detailing how to download this version of the source), can this lead to confusion if a halium-9.0 porter chooses to build the full system image?

abkro (dismissed):
I added a couple of comments about typos or similar. Otherwise I'm good with the changes.

peter-pycom:
done

abkro (dismissed):

abkro (approved):

PR #455, opened 2021-11-01
Remove plasma mobile from docs

kreatoo:
Plasma mobile removed halium support some time ago, so it shouldn't be mentioned in the docs.


istadem2077 (changes requested):
Add Droidian, since it uses Halium too.

doniks:
the remark about droidian might be valid. I'm not familiar with it

however, this doesn't have to be done in this same PR. it looks good as is. I'm merging it

maybe @istadem2077 you could even follow up with a PR about Droidian. In that case it would be helpful for me if you could also point out some reference that shows the relation between Droidian and Halium

doniks (approved):
thanks @kreatoo it's indeed overdue

doniks (approved):

PR #456, opened 2021-11-07
Update index.rst

Ka1z3n:
Would like to take this part out of Devices. Thank you!

### DIFF
--- a/porting/introduction/index.rst
+++ b/porting/introduction/index.rst
@@ -3,7 +3,9 @@ Introduction
 
 Porting is the process of making Ubuntu Touch run on new hardware, i.e. on devices that have not previously been able to run Ubuntu Touch. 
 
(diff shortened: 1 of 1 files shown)


doniks:
Nice! I'm just not convinced about the last sentence. "There is no guarantee..." It seems a little out of place. Also, it strikes me as ... discouraging, or somehow even encouraging to do it only half way ... We have plenty of disclaimers all over the place already. Could we do without that sentence?

Ka1z3n:
Porting is the process of making Ubuntu Touch run on new hardware, i.e. on devices that have not previously been able to run Ubuntu Touch. 

On the vast majority of commercially available devices crucial parts of their systems are not open source. Furthermore, these proprietary parts are specific, not only to the manufacturer, but usually also to the individual model. 

Porting Ubuntu Touch involves building it in such a way as to integrate it with these proprietary components on a device so as to enable it to "talk to" the device hardware.

doniks (approved):
Thanks!

PR #460, opened 2021-11-08
Prior knowledge better place

Ka1z3n:


porting/introduction/index.rst
```diff
@@ -7,6 +7,13 @@ On the vast majority of commercially available devices crucial parts of their sy
 
 Porting Ubuntu Touch involves building it in such a way as to integrate it with these proprietary components on a device so as to enable it to "talk to" the device hardware. 
 
+Prior knowledge and skills
+==========================
```
doniks, line 11: Pls match the length of the underline with the length of the text

doniks (changes requested):

doniks:
I guess we should also remove the text from the old place

doniks:
we discussed that there are three overlapping PRs right now #459 , #460 , #458 . you said you plan to rework those, I'm marking them as drafts for now

PR #478, opened 2021-12-28
Porting corrections

abkro:
Added missing section on console=tty0 for halium-9 plus the correct udev rules command for halium-9 which was previously only found in the porting notes wiki.


doniks (approved):

doniks:
@abkro I was thinking, we could actually merge the h7_build and the h9_build into one build page. what do you think?

abkro:
You may have a point. I'll look through both sections and consider.

PR #481, opened 2022-01-16
links now point to actively developed gitlab repo rather than  archived github repos.

sanathusk:
links now point to actively developed gitlab repos rather than  archived Github repos.


doniks:
CI build fails : https://ci.ubports.com/blue/organizations/jenkins/docs.ubports.com/detail/PR-481/1/pipeline
I think it doesn't like the underscore

One other thing ... I see you updated all po files. I still haven't fully understood how the workflow works between rst/pot/po/weblate ... but all PR's I have seen so far leave the pot&po files alone ... and that seems to work, so, I would be leaning towards sticking with that and not updating those files manually

PR #492, opened 2022-02-28
Rewrite section on overlay files

gber:
This is a complete rewrite of the section on overlay files. It first provides a
separate explanation on how the different overlay mechanisms work and how they
differ between 16.04 and 20.04. After that it explains on how to make use of
that mechanism using the different build methods with an example each.

### DIFF
--- a/porting/configure_test_fix/Overlay.rst
+++ b/porting/configure_test_fix/Overlay.rst
@@ -12,6 +12,15 @@ The way how overlay files are implemented and their limitations vary across buil
 
 In general, there are two methods of implementing overlay files, one is to directly copy them to the destination path potentially overwriting existing files, the other is to place them in a designated base directory from where they are then mounted onto their destination path during the boot process. The former is only possible when doing builds using the GitLab CI scripts. In case of Ubuntu Touch 16.04 it is the only available overlay mechanism. While the ``system.img`` build process allows overwriting files on the built Android system partition it has no means of doing so on the root filesystem. Ubuntu Touch 16.04 does however offer limited support of mount-based overlay files. Starting from Ubuntu Touch 20.04, the implementation of mount-based overlay files has become more powerful in terms of overlaying both files and directories in vendor and Android images regardless of the build method. The use of mount-based overlay files is generally preferable as it allows for delta OTA updates. Both mechanisms will be discussed in detail below before describing how overlay files are included into the build process in full system image builds and builds based on GitLab CI scripts.
 
+.. Note::
+
+    Some ports of Ubuntu Touch 16.04 using the GitLab CI-based build method add custom scripts or configuration files (e.g. in ``/usr/sbin/mount-android.sh``, ``/etc/init/mount-android.conf``, or a file in ``/etc/init/``) in order to overlay files below ``/android/system`` which would otherwise not be possible using this build method. This practice is strongly discouraged for a several reasons:
+
+    - overwriting a system script prevents the port from receiving future updates of that script
+    - different ports coming up with custom solutions creates confusion among new porters and contributors
+
+    Porters which are in need of overlaying files inside a GSI-based ``/android/system.img`` may want to consider using the full ``system.img`` build method instead.
+
 The implementations and limitations can be summarized as follows:
 
 +---------------------------+-------------------------------------------------+--------------------------------------------------------------------------------+
--- a/porting/configure_test_fix/Overlay.rst
+++ b/porting/configure_test_fix/Overlay.rst
@@ -19,20 +19,18 @@ If your port is based on Gitlab CI scripts, it is actually possible to not only 
 Overlay method for full system.img builds
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-In your device directory, create a subdirectory named 'ubuntu'. Collect the files you wish to inject into your build in this directory. 
+In your device directory, create a subdirectory named 'ubuntu'. Collect the files you wish to inject into your build in this directory.
 
 Relevant files are for example (but this list is incomplete):
-    * 70-android.rules
-    * android.conf (for display scaling, see below)
-    * touch.pa (for pulseaudio sound configuration/initialization, see below)
+    * ``/lib/udev/rules.d/70-android.rules``
+    * ``/etc/ubuntu-touch-session.d/android.conf`` (for display scaling, see below)
 
 These files are then injected by adding a code block to the file ``device.mk`` in your device directory. For the three files above add the following code::
 
     ### Ubuntu Touch ###
     PRODUCT_COPY_FILES += \
         $(LOCAL_PATH)/ubuntu/70-android.rules:system/halium/lib/udev/rules.d/70-android.rules \
-        $(LOCAL_PATH)/ubuntu/android.conf:system/halium/etc/ubuntu-touch-session.d/android.conf \
-        $(LOCAL_PATH)/ubuntu/touch.pa:system/halium/etc/pulse/touch.pa 
+        $(LOCAL_PATH)/ubuntu/android.conf:system/halium/etc/ubuntu-touch-session.d/android.conf
     ### End Ubuntu Touch ###
 
 The first of these three files, 70-android.rules, is the one you created when bringing up Lomiri. You can extract this file from the device using ``adb pull`` or ``scp`` and copy it to the 'ubuntu' directory of your device source tree, making sure to add the corresponding line to your ``device.mk`` file, as described above.
@@ -49,7 +47,7 @@ The general steps to follow are thus:
 
 .. Note::
 
-    The target paths for the files mentioned above are *not* randomly chosen. You must use the specified paths. 
+    The target paths for the files mentioned above are *not* randomly chosen. You must use the specified paths.
 
 .. Note::
 
--- a/porting/configure_test_fix/Overlay.rst
+++ b/porting/configure_test_fix/Overlay.rst
@@ -3,91 +3,53 @@
 Overlay file method
 ===================
 
-The UBports rootfs comes with a set of standard configuration files for a number of features such as display scaling, sound, bluetooth and more. These files may not tailored to the needs of your specific device and must therefore to be replaced in order for the feature in question to function as it should. This is done with overlay files, i.e. files that the original files get overwritten with. In other words, you need to rewrite the files in question, making the necessary adjustments for your device, and then see to it that these are incorporated into the build so that they will replace the originals.
+Overview
+--------
 
-If your build is based on the downloaded GSI and rootfs, installed using the halium-install script, then further adjustment of the build will not be possible. This is because the halium-script does not include any method for injecting overlay files. If, on the other hand, your build is based on Gitlab CI scripts, there is a way to further tweak your port with overlay files. This method is described at the end of this section.
+The UBports root filesystem image provides a set of default configuration files for a number of features such as display scaling, sound, bluetooth and more. These files may not be tailored to the device-specific features of the porting target and may therefore have to be adjusted. Furthermore, it may be also be necessary to add additional configuration files as well as scripts for specific needs. Overlay files provide a solution for replacing existing or adding new files to the file system.
 
-.. _Overlay-steps:
+The way how overlay files are implemented and their limitations vary across build methods and different versions of Ubuntu Touch. Use of the halium-install script to install a build based on the downloaded GSI and root filesystem image precludes the use of overlay files since it does not provide a mechanism for that.
 
-Configuring features with overlay files
----------------------------------------
+In general, there are two methods of implementing overlay files, one is to directly copy them to the destination path potentially overwriting existing files, the other is to place them in a designated base directory from where they are then mounted onto their destination path during the boot process. The former is only possible when doing builds using the GitLab CI scripts. In case of Ubuntu Touch 16.04 it is the only available overlay mechanism. While the ``system.img`` build process allows overwriting files on the built Android system partition it has no means of doing so on the root filesystem. Ubuntu Touch 16.04 does however offer limited support of mount-based overlay files. Starting from Ubuntu Touch 20.04, the implementation of mount-based overlay files has become more powerful in terms of overlaying both files and directories in vendor and Android images regardless of the build method. The use of mount-based overlay files is generally preferable as it allows for delta OTA updates. Both mechanisms will be discussed in detail below before describing how overlay files are included into the build process in full system image builds and builds based on GitLab CI scripts.
 
-For full ``system.img``-based builds, irrespective of Halium version, the method described below applies if the file you need to edit can be found in the /etc directory (or a subdirectory of this) on your device. In other words, the file or files you wish to overlay must actually exist and be located where specified above. You should not attempt to overwrite files located elsewhere using the method described here.
+The implementations and limitations can be summarized as follows:
 
-If your port is based on Gitlab CI scripts, it is actually possible to not only overlay existing files, but also introduce files that were not originally present. The method for Gitlab CI script based builds is detailed at the end of this section. If this is the method you will be using, you should still read both sections to gain a better understanding of how this method actually works.
++---------------------------+-------------------------------------------------+--------------------------------------------------------------------------------+
+| Build method              | Ubuntu Touch 16.04                              | Ubuntu Touch 20.04                                                             |
++===========================+=================================================+================================================================================+
+| full ``system.img`` build | - add/overwrite files below ``/android/system`` | - add/overwrite files below ``/android/system``                                |
+|                           | - overlay existing files on the root filesystem | - overlay existing files on the root filesystem, android and vendor filesystem |
+|                           | - no new files on the root filesystem           | - new files and directories via merged or replaced overlay directories         |
++---------------------------+-------------------------------------------------+--------------------------------------------------------------------------------+
+| GitLab CI-based build     | - add/overwrite files on the root file system   | - add/overwrite files on the root file system (use overlays instead)           |
+|                           |                                                 | - overlay mounts for existing files on the root, android and vendor filesystem |
+|                           |                                                 | - new files and directories via merged or replaced overlay directories         |
++---------------------------+-------------------------------------------------+--------------------------------------------------------------------------------+
 
-Overlay method for full system.img builds
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-In your device directory, create a subdirectory named 'ubuntu'. Collect the files you wish to inject into your build in this directory.
-
-Relevant files are for example (but this list is incomplete):
-    * ``/lib/udev/rules.d/70-android.rules``
-    * ``/etc/ubuntu-touch-session.d/android.conf`` (for display scaling, see below)
-
-These files are then injected by adding a code block to the file ``device.mk`` in your device directory. For the three files above add the following code::
-
-    ### Ubuntu Touch ###
-    PRODUCT_COPY_FILES += \
-        $(LOCAL_PATH)/ubuntu/70-android.rules:system/halium/lib/udev/rules.d/70-android.rules \
-        $(LOCAL_PATH)/ubuntu/android.conf:system/halium/etc/ubuntu-touch-session.d/android.conf
-    ### End Ubuntu Touch ###
-
-The first of these three files, 70-android.rules, is the one you created when bringing up Lomiri (see :ref:`Lomiri`). Copy it to the 'ubuntu' directory of your device source tree, making sure to add the corresponding line to your ``device.mk`` file, as described above.
-
-Explanation:
-
-The string before the colon '$(LOCAL_PATH)/ubuntu/70-android.rules' specifies the path to the source file to be injected. The string after the colon 'system/halium/lib/udev/rules.d/70-android.rules' specifies the target location on your device.
-
-The general steps to follow are thus:
-    1. Copy the file you wish to modify to the 'ubuntu' directory you have created in your device source tree.
-    2. Edit the file as needed.
-    3. Add a line to the PRODUCT_COPY_FILES section of your device.mk file as shown above.
-    4. Rebuild your system.img and reflash together with the ubports rootfs.
-
-.. Note::
-
-    The target paths for the files mentioned above are *not* randomly chosen. You must use the specified paths.
-
-.. Note::
-
-    When you specify target path 'system/halium/etc/myfilename' your file 'myfilename' will end up in the '/etc' directory of your device (i.e. without the leading 'system/halium')
-
-.. _Rebuild-system.img:
-
-Rebuild system.img
-""""""""""""""""""
-
-For Halium-7.1 ports and Halium-9.0 ports using a device specific system.img in place of the GSI, the system.img now needs to be rebuilt.
-
-When you have made the adjustments you need and prepared your source as described above, rebuild the system image: ``mka systemimage``. 
-
-When rebuilding the system image after small changes like these, you need not ``mka clean`` first. However, changes to PRODUCT_PROPERTY_OVERRIDES might not get detected by the build system. To ensure this, go to the directory ``BUILDDIR/out/target/product/CODENAME/system`` and delete the file ``build.prop`` before rebuilding.
-
-Overlay method for Gitlab CI script-based builds
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Builds in this category can also be adapted to a particular device using overlay files, and when building by this method (see :ref:`Gitlab-CI`) it is possible not only to replace existing files, but also to introduce new ones.
+How mount-based overlay files work
+----------------------------------
 
 Ubuntu Touch 16.04
-""""""""""""""""""
+^^^^^^^^^^^^^^^^^^
 
-Ubuntu Touch 16.04 only allows replacing or introducing individual files on the Ubuntu Touch root filesystem. All of those overlay files need to be placed in the GitLab repository replicating their paths relative to the directory ``overlay/system`` before running the Gitlab CI scripts according to `the instructions included <https://gitlab.com/ubports/community-ports/android9/xiaomi-redmi-note-7-pro/xiaomi-violet/-/blob/master/README.md>`_.
+Ubuntu Touch 16.04 only offers mount-based overlay files in case of full ``system.img``-based builds with the limitation that the overlayed files must actually exist on the root filesystem partition, adding new files is not possible.
 
-For example, in order to introduce the udev rules file ``70-android.rules`` created in the last step in  ``/lib/udev/rules.d/`` on the system it needs to be placed in the directory ``overlay/system/halium/lib/udev/rules.d/`` in the repository.  The build scripts will include it in the build and the overlay files will be bind-mounted onto their final locations during the boot process.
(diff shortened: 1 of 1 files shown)


porting/configure_test_fix/Overlay.rst
```diff
@@ -3,82 +3,133 @@
...
 
-The first of these three files, 70-android.rules, is the one you created when bringing up Lomiri. You can extract this file from the device using ``adb pull`` or ``scp`` and copy it to the 'ubuntu' directory of your device source tree, making sure to add the corresponding line to your ``device.mk`` file, as described above.
+- If a subdirectory contains a special file named ``.halium-override-dir`` that directory will replace the destination directory tree, i.e. the contents of the underlying directory and its descendants will no longer be accessible.
 
-Explanation:
+- In case a special file named ``.halium-overlay-dir`` exist in a subdirectory, it will be merged with with the destination directory.  Any file in the underlying directory or descendants thereof remain accessible if no file with the same destination path exists in the overlaying directory tree.  Files and whole directories which only exist in the overlay are made accessible in their respective destination paths.  This is implemented using the overlayfs filesystem (see the `overlayfs documentation <https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html>`_ for the technical details).
```
peat-psuwit, line 33: > If a subdirectory contains a special file...

> In case a special file named...

Consistency?

porting/configure_test_fix/Overlay.rst
```diff
@@ -3,82 +3,133 @@
...
+The actual overlay mechanism for replacing or adding files depends both on the build method and the version of Ubuntu Touch. Using the halium-install script to install a build based on the downloaded GSI and rootfs precludes the use of overlay files since it does not provide a mechanism for that.
 
-.. _Overlay-steps:
+In general, there are two methods of implementing overlay files, one is to directly extract them in the target path potentially overwriting existing files, the other is to place them in a designated path from where they are then mounted onto their destination path during the boot process. The latter method is generally preferable as it allows for delta OTA updates and, starting from Ubuntu Touch 20.04, has become more flexible in terms of overlaying files and directories in vendor and Android images. It will be discussed in detail below before describing how overlay files are included into the build process in full system image builds and builds based on GitLab CI scripts.
```
peat-psuwit, line 10: > one is to directly extract them in the target path potentially overwriting existing files,...

I think the documentation is not clear yet about what it means to "directly extract".

It's in the format of the device tarball. Anything in `system/` directory in the tarball [1] will be directly extracted into the on-device rootfs by `system-image-upgrader`.

If this device tarball is built as part of full-system.img port, this portion of the device tarball will contain only `system.img` or `android-rootfs.img` (see [here](https://gitlab.com/ubports/community-ports/jenkins-ci/halium-build-tools/-/blob/main/build-tarball.sh)), and no "directly extracted" overlay file exists.

If, however, it's built as part of "Gitlab CI script-based builds", this directory can also include "directly extracted" overlay files (see [here](https://gitlab.com/ubports/community-ports/android9/volla-phone-x/volla-yggdrasilx/-/blob/master/build.sh#L86-87)). This also means, without modifying the build script, placing files in `overlay/` directory of the git repo will _actually_ make them "directly extracted" overlay, which is currently the case for all GItLab CI script-based ports since mount-based system for them is not available for 16.04. [2]

[1] Not to be confused with Android's `/system`. Oh, I hate this naming...
[2] This is also worries me, as the tarball build scripts stays with each device, which means they're _tens of forks_ of these scripts. I have a feeling that we might have to edit them all...

porting/configure_test_fix/Overlay.rst
```diff
@@ -3,82 +3,133 @@
...
 
-For example:
-""""""""""""
+Example
+"""""""
 
-The udev rules file ``70-android.rules`` needs to go into ``system/halium/lib/udev/rules.d/``. This is accomplished by first creating the directory in the Gitlab scripts build tree. Standing in the root of this tree, first create the directory and then copy the file to this location::
+Following the above example, adding the overlay files ``/lib/udev/rules.d/70-android.rules`` and ``/etc/ubuntu-touch-session.d/android.conf`` requires them to end up below ``/opt/halium-overlay`` on the resulting filesystem, that is as ``/opt/halium-overlay/lib/udev/rules.d/70-android.rules`` and ``/opt/halium-overlay/etc/ubuntu-touch-session.d/android.conf``. This is achieved by placing them as ``overlay/system/lib/udev/rules.d/70-android.rules`` and ``overlay/system/etc/ubuntu-touch-session.d/android.conf`` in the repository.
```
peat-psuwit, line 130: > This is achieved by placing them as ``overlay/system/lib/udev/rules.d/70-android.rules``...

As I discussed earlier, this is incorrect. At least without a modification to the build script...

I just realised that, to use mount-based overlay, one has to actually do "directly extracted" overlay to `/opt/halium-overlay` first. How can we eliminate this confusion...

Like, as it currently stands, porters can choose to use "directly extracted" overlay by placing files directly on `overlay/system`, or choose to use "mount-based overlay" by "directly extracted" overlay over `/opt/halium-overlay` i.e. placing files at `overlay/system/opt/halium-overlay`.

peat-psuwit (changes requested):

peat-psuwit:
@NotKit could you please weigh-in from the porters' perspective?

porting/configure_test_fix/Overlay.rst
gber: > If this device tarball is built as part of full-system.img port, this portion of the device tarball will contain only system.img or android-rootfs.img (see [here](https://gitlab.com/ubports/community-ports/jenkins-ci/halium-build-tools/-/blob/main/build-tarball.sh)), and no "directly extracted" overlay file exists.

But in this case the `PRODUCT_COPY_FILES` macro will do the equivalent to "extraction" by copying it onto the filesystem image, will it not?

porting/configure_test_fix/Overlay.rst
gber: Corrected for the example. Do you think there we need a second example showing directly extracted files for 16.04?

porting/configure_test_fix/Overlay.rst
peat-psuwit: Unfortunately, no. We're in Halium/Android build system, and it only produces `system.img`/`android-rootfs.img` (and `halium-boot.img`).

Output paths in `PRODUCT_COPY_FILES` are relative to Android's `$OUT` directory. Files in `$OUT/system/` will be in `/` of `system.img` or in `/system/` of `android-rootfs.img`, both of which results in files being available at `/android/system` after Android image mounting is done.

As I said, don't be confused with device tarball's `/system/` which will be extracted over the root partition on device. As noted earlier, for `system.img` ports the device tarball is created in a step after that.

gber: OK, thanks for the detailed explanation (again). I've made corrections accordingly.

What I'm still unclear about is when using the Gitlab CI scripts for xenial: Is there a way to use bind mounts? `/opt/halium-overlay` is only supported starting from your modifications in focal, right? So is the only option to overlay files in xenial by extraction?

peat-psuwit: No, Gitlab CI script ports cannot use bind mounts. On xenial, the overlay mounting code looks only in `/android/system`, and since these devices have the same `android-rootfs.img`, that cannot be done.

So, yeah, on xenial they'll have to stick to the directly extracted method.

porting/configure_test_fix/Overlay.rst
```diff
@@ -3,82 +3,150 @@
...
+|                           | - no new files on the root filesystem           | - new files and directories via merged or replaced overlay directories         |
++---------------------------+-------------------------------------------------+--------------------------------------------------------------------------------+
+| GitLab CI-based build     | - add/overwrite files on the root file system   | - add/overwrite files on the root file system (use overlays instead)           |
+|                           |                                                 | - overlay mounts for existing files on the root, android and vendor filesystem |
+|                           |                                                 | - new files and directories via merged or replaced overlay directories         |
++---------------------------+-------------------------------------------------+--------------------------------------------------------------------------------+
 
-Overlay method for full system.img builds
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Mount-based overlay files
```
peat-psuwit, line 29:
```suggestion
Understand mount-based overlay files
```

?

porting/configure_test_fix/Overlay.rst
```diff
@@ -3,82 +3,150 @@
...
+    └── quuxrc (replaced)
 
-    The target paths for the files mentioned above are *not* randomly chosen. You must use the specified paths. 
+The directory ``/etc/foo`` has been merged, whereas the ``/etc/bar`` directory and the file ``/etc/quuxrc`` have been replaced.
 
-.. Note::
+.. _Overlay-steps:
+
+Overlay files in full system.img builds
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
peat-psuwit, line 111:
```suggestion
------------------------------------------
```

To make the heading the same size as "Understading mount-based overlay".

porting/configure_test_fix/Overlay.rst
```diff
@@ -3,82 +3,150 @@
...
 
-Overlay method for Gitlab CI script-based builds
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+After a successful build, the resulting ``system.img`` must reflashed together with the UBports root filesystem. On boot files and – in case of Ubuntu Touch 20.04 or later – marked directories in ``/android/system/halium`` will be mounted onto their final locations which are derived from their path relative to the designated base directory.
 
-Builds in this category can also be adapted to a particular device using overlay files, and when building by this method (see :ref:`Gitlab-CI`) it is possible not only to replace existing files, but also to introduce new ones.
+Overlay files in Gitlab CI script-based builds
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
peat-psuwit, line 140: Ditto.

porting/configure_test_fix/Overlay.rst
```diff
@@ -3,82 +3,150 @@
...
 
-When you have prepared the files to replace or introduce, and determined the exact locations where they need to go on your device, these locations need to be replicated under the directory ``overlay/system/`` before running the Gitlab CI scripts according to `the instructions included. <https://gitlab.com/ubports/community-ports/android9/xiaomi-redmi-note-7-pro/xiaomi-violet/-/blob/master/README.md>`_
+When using Gitlab CI script-based builds (see :ref:`Gitlab-CI`) overlay files and directories have to be placed in in the repository below ``overlay/system`` in a directory tree mirroring the structure of the root filesystem. Note that any overlay file placed in that directory tree will overwrite an existing file on the root filesystem. While Ubuntu Touch 16.04 only allows overlaying files by overwriting them, in case of Ubuntu Touch 20.04 or later versions the mount-based overlay files should be used instead. Thus overlay files and directories should be placed in a directory tree below ``overlay/system/opt/halium-overlay`` in the repository rather than directly below ``overlay/system``.
```
peat-psuwit, line 142: I wonder if we should add a reference about device tarball here... However, I'm not sure how it should be added.

porting/configure_test_fix/Overlay.rst
```diff
@@ -3,82 +3,150 @@
...
+Example
+"""""""
 
-The udev rules file ``70-android.rules`` needs to go into ``system/halium/lib/udev/rules.d/``. This is accomplished by first creating the directory in the Gitlab scripts build tree. Standing in the root of this tree, first create the directory and then copy the file to this location::
+Following the above example on Ubuntu Touch 20.04, adding the overlay files ``/lib/udev/rules.d/70-android.rules`` and ``/etc/ubuntu-touch-session.d/android.conf`` requires them to end up below ``/opt/halium-overlay`` on the resulting filesystem, that is as ``/opt/halium-overlay/lib/udev/rules.d/70-android.rules`` and ``/opt/halium-overlay/etc/ubuntu-touch-session.d/android.conf``. This is achieved by placing them at ``overlay/system/opt/halium-overlay/lib/udev/rules.d/70-android.rules`` and ``overlay/system/opt/halium-overlay/etc/ubuntu-touch-session.d/android.conf`` in the repository. In case of Ubuntu Touch 16.04 the files on the root filesystem need to be overwritten due to the lack of mount-based overlay files and must thus be placed at ``overlay/system/lib/udev/rules.d/70-android.rules`` and ``overlay/system/etc/ubuntu-touch-session.d/android.conf`` in the repository.
```
peat-psuwit, line 147: Maybe add a line break at "In case of Ubuntu Touch 16.04..."?

peat-psuwit (changes requested):

peat-psuwit:
BTW, Step Alex on Telegram suggested that we also mention overriding `mount-android.sh` to get things overlaid in `/android/system` in 16.04. However, I don't particularly like this method, since it overrides core script, blocking that port from receiving update for that script.

How should we mention it? Just briefly talks about it and then suggesting people that need to override stuff in `/android/system` to "consider doing full system.img port"?

porting/configure_test_fix/Overlay.rst
gber: "How mount-based overlay files work"?

porting/configure_test_fix/Overlay.rst
gber: OK, I've moved both sub-subsections up one level.

porting/configure_test_fix/Overlay.rst
gber: See above.

porting/configure_test_fix/Overlay.rst
gber: OK.

porting/configure_test_fix/Overlay.rst
gber: My thinking was that this is an internal implementation detail that a porter does not strictly have to know about. We could add something like:

> The contents of `overlay/system` are included as the directory `system` in the device tarball and copied onto the root filesystem during the installation or upgrade process by system-image-upgrader.

Your call.

gber:
> BTW, Step Alex on Telegram suggested that we also mention overriding `mount-android.sh` to get things overlaid in `/android/system` in 16.04. However, I don't particularly like this method, since it overrides core script, blocking that port from receiving update for that script.
> 
> How should we mention it? Just briefly talks about it and then suggesting people that need to override stuff in `/android/system` to "consider doing full system.img port"?

Yes, I think that's a reasonable approach. I've added a small note describing and discouraging the practice and outlining alternatives.

porting/configure_test_fix/Overlay.rst
peat-psuwit: The second example is added in the new version.

porting/configure_test_fix/Overlay.rst
peat-psuwit: Hmm ok. It could be mentioned in another documentation page/section about Ubuntu Touch's internal.

nitanmarcel:
> BTW, Step Alex on Telegram suggested that we also mention overriding `mount-android.sh` to get things overlaid in `/android/system` in 16.04. However, I don't particularly like this method, since it overrides core script, blocking that port from receiving update for that script.
> 
> How should we mention it? Just briefly talks about it and then suggesting people that need to override stuff in `/android/system` to "consider doing full system.img port"?

Did I? Probably I forgot or there was a miss-communication issue here.

Anyway. `mount-android.sh` is a critical thing needed for the device to successfully mount android partitions and make the device boot. Which seems like it's a generic issue which is yet not fixed in the rootfs itself or probably it is but we haven't test to see if removing the overlay affects the booting process in any way.

About using it the override the android files:

As seen [here](https://gitlab.com/ubports/community-ports/android10/xiaomi-redmi-note-9-pro/xiaomi-miatoll/-/blob/halium-10.0/overlay/system/etc/init/apply-halium-overlay.conf) his can be implemented in a separate init script without overriding the system's mount-android file. It's up to the porter to decide.

gber:
> @gber, should the wording in the note be changed?

As in and additional note "don't add random upstart jobs, use a different porting method"?

luksus42:
The overlay, as described in the docs does not seem to work for me in the Gitlab CI build for FP3.
Putting the overlay in `/overlay/system/opt/halium-overlay/` makes the device not booting in any accessible state like telnet or ssh.
Just moving the `lib` overlay in to /overlay/system/usr/ seems to work currently, as done here: https://gitlab.com/ubports/community-ports/android10/fairphone/fairphone_fp3/-/tree/halium-10.0-focal/overlay/system.

I can connect through ssh-over-usb. But no display is coming up. There are lots of issues regarding init, due to a read-only system, like:
```
[   12.842678] init: starting service 'vendor.per_mgr'...
[   12.850638] init: createProcessGroup(1000, 234) failed for service 'vendor.per_mgr': Read-only file system
[   12.853598] init: starting service 'ashmemd'...
[   12.860512] init: createProcessGroup(9999, 235) failed for service 'ashmemd': Read-only file system
[   12.869961] init: starting service 'gpu'...
[   12.875853] init: createProcessGroup(1072, 236) failed for service 'gpu': Read-only file system
[   12.882132] init: starting service 'vendor.rmt_storage'...
[   12.886428] init: createProcessGroup(0, 239) failed for service 'vendor.rmt_storage': Read-only file system
[   12.901222] init: starting service 'vendor.tftp_server'...
[   12.920066] init: createProcessGroup(0, 241) failed for service 'vendor.tftp_server': Read-only file system
[   12.921770] init: Command 'class_start core' action=boot (/init.rc:779) took 133ms and succeeded
[   12.921895] init: processing action (boot) from (/vendor/etc/init/hw/init.qcom.rc:112)
[   12.928831] init: Service 'irsc_util' (pid 230) exited with status 0
[   12.939334] init: Command 'mkdir /persist/drm 0770 system system' action=boot (/vendor/etc/init/hw/init.qcom.rc:151) took 0ms and failed: mkdir() failed: Read-only file system
[   12.939499] init: Received control message 'interface_start' for 'android.hardware.configstore@1.0::ISurfaceFlingerConfigs/default' from pid: 25 (/system/bin/hwservicemanager)
[   12.939559] init: Could not find 'android.hardware.configstore@1.0::ISurfaceFlingerConfigs/default' for ctl.interface_start
[   12.940734] init: Command 'mkdir /persist/misc 0770 system system' action=boot (/vendor/etc/init/hw/init.qcom.rc:155) took 0ms and failed: mkdir() failed: Read-only file system
[   12.943743] init: Command 'mkdir /persist/alarm 0770 system system' action=boot (/vendor/etc/init/hw/init.qcom.rc:156) took 1ms and failed: mkdir() failed: Read-only file system
[   12.951334] init: Received control message 'interface_start' for 'vendor.display.config@1.7::IDisplayConfig/default' from pid: 25 (/system/bin/hwservicemanager)
[   12.951385] init: Could not find 'vendor.display.config@1.7::IDisplayConfig/default' for ctl.interface_start
[   12.952492] init: Received control message 'interface_start' for 'vendor.display.config@1.7::IDisplayConfig/default' from pid: 25 (/system/bin/hwservicemanager)
[   12.952538] init: Could not find 'vendor.display.config@1.7::IDisplayConfig/default' for ctl.interface_start
[   12.953726] init: Command 'chmod 0755 /system/bin/ip' action=boot (/vendor/etc/init/hw/init.qcom.rc:170) took 0ms and failed: fchmodat() failed: Read-only file system
[   12.965652] init: Command 'mkdir /persist/display 0770 system graphics' action=boot (/vendor/etc/init/hw/init.qcom.rc:226) took 0ms and failed: mkdir() failed: Read-only file system
```

I had a successful build one month ago, where display did come up with the lxc-patches from Alexander Martinz (https://bugs.launchpad.net/ubuntu/+source/lxc/+bug/1958661).
But this is not working anymore...

peat-psuwit:
@luksus42 ok, interesting...

When there's no connectivity whatsoever, one way that could recover the log is to reboot to recovery and look at `/data/system-data/var/log/journal`. You may have to pull that to your machine, and then point `journalctl` to it.

Other than that... I'm really not sure what's going on. The bind-mount code is shared between the system.img case and `/opt/halium-overlay` case, and on Volla Phone it seems to work. You have configured `.halium-{override,overlay}-dir`, right?

Maybe I really need another phone with GitLab CI-based port...

luksus42:
> You have configured `.halium-{override,overlay}-dir`, right?

Uhm, no. I thought this would be kind of optional, and I forgot about it afterwards... will try it again.
Can you point me to the Volla Repo, where the changes are already done? Or isn't it public yet?

peat-psuwit:
This is what I've done to overlay `/android/vendor` on Volla Phone. It's not merged yet, though...

https://github.com/HelloVolla/android_device_volla_yggdrasil/pull/13/files#diff-247ee86229a709a6e2eedfc1d3c4a557825aee073e20b0112ab76f4ca8e4bc4eR79

Although, admittedly the overlay in Volla Phone is not big, and maybe not the best example.

> I thought this would be kind of optional, and I forgot about it afterwards...

Well, it's not optional in a couple of areas (e.g. `/lib/modules` where you need exact directory name, or when you add new units in `/lib/systemd/system`).

For your case, maybe it's because `70-fp3.rules` is not named as `70-android.rules`, and thus doesn't get bind mounted. Either place `.halium-overlay-dir` in `/lib/udev/rules.d` or, more preferably, just name the file `70-android.rules` instead.

luksus42:
I tried first with renaming the rules file, second with putting .halium-overlay-dir file in every directory.
Result is the same for both:
- the first boot gives me access to ssh (now without write-permission errors, it seems)
- the second boot gives me nothing, except of the usb device authenticating as `Bus 001 Device 023: ID 1209:0002 Generic pid.codes Test PID` when querying with lsusb. If I remember correctly, this happened also on my first attempt without any changes...
  - maybe I should remove (or move?) the usb-mtp overlay for now
- installing the lxc patches makes the display go black

peat-psuwit:
@gber

> As in an additional note "don't add random upstart jobs, use a different porting method"?

Hmm... not exactly... I'm thinking something like "(or a similar method which add codes to overlay files in `/android/system`)".

peat-psuwit:
@luksus42 

> I tried first with renaming the rules file, second with putting .halium-overlay-dir file in every directory. Result is the same for both:

Hmm... actually `.halium-overlay-dir` is not actually supposed to be used in every directory. However I don't have a good answer yet about when & where it should be used. The problem is, I don't have a good visibility what porters do nowadays, and I'm not sure what I should do...

> * the first boot gives me access to ssh (now without write-permission errors, it seems)
> * the second boot gives me nothing, except of the usb device authenticating as `Bus 001 Device 023: ID 1209:0002 Generic pid.codes Test PID` when querying with lsusb. If I remember correctly, this happened also on my first attempt without any changes...
>   
>   * maybe I should remove (or move?) the usb-mtp overlay for now

The usb-mtp overlay, given that it's on upstart, shouldn't do anything for now. We have the new system for USB mode management for Focal [1]. The behavior should be that it should consistently give you network interface if you keep the cable plugged in during boot.

For now, `1209:0002` means MTP (which requires Lomiri to be up), while `1209:0004` means RNDIS.

[1] For a sneak peak, https://gitlab.com/ubports/core/packaging/usb-moded/-/wikis/home

> * installing the lxc patches makes the display go black

Here a full `journalctl` log would be helpful. Also, you may be interested in [2].

[2] https://bugs.launchpad.net/ubuntu/+source/lxc/+bug/1959993/comments/15

gber:
> @gber
> 
> > As in an additional note "don't add random upstart jobs, use a different porting method"?
> 
> Hmm... not exactly... I'm thinking something like "(or a similar method which add codes to overlay files in `/android/system`)".

Could you elaborate a bit what whole the content of the note should be?

peat-psuwit:
@gber

> Could you elaborate a bit what whole the content of the note should be?

Err... sorry. Let's try:

> Some ports of Ubuntu Touch 16.04 use the GitLab CI-based build method to add a custom code (e.g. in /usr/sbin/mount-android.sh, /etc/init/mount-android.conf, or a file unter /etc/init/) to enable overlayingy files below /android/system (which would otherwise not be possible using this build method). This practice is strongly discouraged for a several reasons:

> - overwriting a system script prevents the port from receiving future updates of that script.
> - Adding custom code make means different ports could use different methods, which could be confusing for new porters. <Guido: some other reason?>

> If you've found yourself in need of overlaying files inside GSI-based /android/system.img, you may want to consider using full-system.img method instead.

gber:
Allright, I've incorporated that with some minor language/stylistic modifications.

doniks:
I didn't read the whole thread, but allow me to ask @peat-psuwit @gber where do we stand with this?

peat-psuwit:
@doniks Ah, sorry. I intended to try to follow this doc as part of a port migration, but never get around to do that.

I think this doc is technically correct now. Whether it includes all required information, or if it's easy to understand, I would like you @doniks to review. If you think this looks OK, you can merge it.

A preview of this PR's page is at:

https://ci.ubports.com/job/docs.ubports.com/job/PR-492/lastSuccessfulBuild/artifact/_build/html/porting/configure_test_fix/Overlay.html

peat-psuwit (approved):

doniks:
mhm ... just one point ... maybe it's just me not getting it though, or a typo ... 

In the [example](https://ci.ubports.com/job/docs.ubports.com/job/PR-492/lastSuccessfulBuild/artifact/_build/html/porting/configure_test_fix/Overlay.html#example) 
there is a bazrc file in the overlay, but it doesn't show up in the result. From my (limited) understanding I would have expected it. Also the quuxrc is shown as (replaced) in the result, but I would have thought it is there unchanged taken from the image

peat-psuwit:
Nice catch!

> In the [example](https://ci.ubports.com/job/docs.ubports.com/job/PR-492/lastSuccessfulBuild/artifact/_build/html/porting/configure_test_fix/Overlay.html#example) there is a `bazrc` file in the overlay, but it doesn't show up in the result. From my (limited) understanding I would have expected it.

This is because `bazrc` doesn't exists in the rootfs but in the overlay, and the porter doesn't configure the directory `bazrc` is in to be in "merge dir" mode. I think this would require explanation. @gber ?

> Also the `quuxrc` is shown as (replaced) in the result, but I would have thought it is there unchanged taken from the image

I think this is the typo. This file should appear in the rootfs isn't it @gber?

And finally, I realized that `.halium-overlay-dir` shouldn't be in the rootfs at all. So it should be removed from the rootfs part of the example.

gber:
> Nice catch!
> 
> > In the [example](https://ci.ubports.com/job/docs.ubports.com/job/PR-492/lastSuccessfulBuild/artifact/_build/html/porting/configure_test_fix/Overlay.html#example) there is a `bazrc` file in the overlay, but it doesn't show up in the result. From my (limited) understanding I would have expected it.
> 
> This is because `bazrc` doesn't exists in the rootfs but in the overlay, and the porter doesn't configure the directory `bazrc` is in to be in "merge dir" mode. I think this would require explanation. @gber ?
> 
> > Also the `quuxrc` is shown as (replaced) in the result, but I would have thought it is there unchanged taken from the image
> 
> I think this is the typo. This file should appear in the rootfs isn't it @gber?

Right, `quuxrc` should in fact be named `bazrc`, the point is to illustrate each of the three ways files and directories may be overlaid.

> And finally, I realized that `.halium-overlay-dir` shouldn't be in the rootfs at all. So it should be removed from the rootfs part of the example.

Indeed, a copy & paste error.

doniks:
> > And finally, I realized that `.halium-overlay-dir` shouldn't be in the rootfs at all. So it should be removed from the rootfs part of the example.
> 
> Indeed, a copy & paste error.

`.halium-overlay-dir` is still there. Also `.halium-override-dir` is there. I'm a little lost on whether that's correct as is or not. I don't think it's super critical, so I leave my approve already

doniks (approved):

PR #501, opened 2022-05-25
Update dav.rst

Marc-Flender:
Hello, 
I propose an add-on (DBUS_SESSION_BUS_ADDRESS=\$) because cron jobs don't work without this correction.

Nevertheless, i think a better script is present here
https://github.com/UT-ilities/UTouch_CalDAV_and_CardDAV_synchronization
So maybe you should modify completely te page and link to this one?
Don't you think?


Marc-Flender:
Hello @doniks 
As a matter of fact, bastos77 modified the script, so it resolves the issue. 

I will propose a change for using various options using the other script

PR #518, opened 2022-08-17
Update webapp guide

maciek134:
Change http to https in links, fix some semantics and outdated/incorrect information.

### DIFF
--- a/appdev/webapp/guide.rst
+++ b/appdev/webapp/guide.rst
@@ -9,8 +9,7 @@ A web app displays in a webview inside a webapp-container that runs as an Ubuntu
 Taking a closer look:
 
     At the innermost level, there is a website that the developer identifies by URL.
-    The website is rendered and runs in an Oxide webview. Oxide is a Blink/Chrome webview that is customized for Ubuntu.
-    The Oxide webview runs and displays in the webapp-container. The webapp-container is the executable app runtime that is integrated with the Ubuntu/unity shell.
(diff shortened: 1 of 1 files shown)


appdev/webapp/guide.rst
```diff
@@ -164,9 +163,9 @@ To change the string from the command line, use the following option:::
...
 ------------------------
 
-The webapp experience is contained and isolated from the browser data point of view. That is webapps do not access data from any other installed browser, such as history, cookies and so on. Other browser on the system do not access the webapp’s data.
+The webapp experience is contained and isolated from the browser data point of view. That is webapps do not access data from any other installed browser, such as history, cookies and so on. Other browsers on the system do not access the webapp’s data.
 
 Storage
 -------
 
-W3C allows apps to use local storage, and Oxide/Webapp-container supports the main standards here: LocalStorage, IndexedDB, WebSQL.
+Since webapp-container is based on Chromium it supports some of the same storage options (Local/Session Storage, IndexedDB and the deprecated WebSQL). FileSystem API is not supported because apps cannot access the filesystem directly.
```
doniks, line 171: I'm not an expert on this, but it seems we previously explained that these things are generic standards that we also support. The new formulation makes it sound like it's proprietary Chromium features. I think this is not the case 

https://en.wikipedia.org/wiki/Web_storage#localStorage

How about this:

"Webapp-container supports the following W3C storage options: Local/Session Storage, IndexedDB and the deprecated WebSQL. The FileSystem API is not supported ...."

doniks (changes requested):

appdev/webapp/guide.rst
maciek134: Hmm, I guess it does ;p fixing

doniks (approved):
thanks!

PR #525, opened 2022-10-04
Move the reboot step after enabling developer mode

maciek134:
The device has to be rebooted after developer mode is enabled, rebooting it earlier doesn't do anything.


JamiKettunen:
+1 from me, @doniks

doniks (approved):

PR #497, opened 2022-05-03
porting: finalize: add instructions for official UBports support

amartinz:
Requirements for a port to get officially supported by UBports.

PR #540, opened 2023-01-10
porting: finalize: add instructions for official UBports support

doniks:
Requirements for a port to get officially supported by UBports.

Change-Id: Ibe53407ec2909b4842eacf07e87d5cccac8389cc
Signed-off-by: Alexander Martinz <amartinz@shiftphones.com>

### DIFF
--- a/porting/finalize/UBports-official-support/index.rst
+++ b/porting/finalize/UBports-official-support/index.rst
@@ -1,0 +1,87 @@
+.. _OfficialSupport:
+
+Get your port officially supported
+==================================
+
+Having your port officially supported comes with certain benefits, such as:
+
+* Official UBports installer support to allow easier installation of your port
+* Port system images are getting built via our CI infrastructure and delivered as OTA updates
+* Daily built *devel* images, which allows for easy testing of the latest state of Ubuntu Touch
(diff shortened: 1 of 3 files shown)


doniks:
ok. I've fixed the rst formatting, so from a *layout* and code formatting etc perspective I'm fine with it. 

From a *content* perspective, this https://github.com/ubports/docs.ubports.com/pull/497#discussion_r975714881 would still be a question form me. 

@fredldotme can you please have a look

doniks:
@Flohack74 to followup on what you said in the other PR https://github.com/ubports/docs.ubports.com/pull/497

> @doniks the problem is: Even if only a handful of people need to agree on this I do not want to push this entirely on my own and with Alex. I would have merged this already, but there is little interest in the porting corner to get this moving I guess. It is also unclear who are the stakeholders, and how we empower them to execute all of this.
> 
> If we get more restrictive then we should really be serious. It needs to be a logical step forward, but sometimes I have the feeling it is not what people want when saying "mobile freedom". Or idk.

Okay. Fair enough. If there is no consensus among the relevant people, then it doesn't make sense to write this up. Thanks for letting me know!

PR #545, opened 2023-01-22
Update Preparations.rst

ZGStuff:
just updated version numbers for ubuntu

### DIFF
--- a/porting/introduction/Preparations.rst
+++ b/porting/introduction/Preparations.rst
@@ -40,6 +40,6 @@ You need to install a custom recovery image on your device in order to handle fo
 Recommendations for the host/build PC
 -------------------------------------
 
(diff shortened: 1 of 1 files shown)


ZGStuff:
dang MORE updates? another one pull request it is

doniks (changes requested):
I think it might be best to remove the whole "For example" sentence. 

It doesn't really convey any information. Especially since there is the other place here https://github.com/ubports/docs.ubports.com/pull/546. As it is written now it is bound to be regularly out of date.

doniks (changes requested):
I think it might be best to remove the whole "For example" sentence.

Flohack74:
Also , please make sure to run `update-translations.sh` on top level of the checkout with each change in an rst file, otherwise we cannot get the pot files updated and nothing can be translated :) - Instructions how to set up sphinx and other stuff should be in the README

doniks:
> Also , please make sure to run `update-translations.sh` on top level of the checkout with each change in an rst file, otherwise we cannot get the pot files updated and nothing can be translated :) - Instructions how to set up sphinx and other stuff should be in the README

sounds like generic instructions?! Or is this specific for this PR? I'm worried if we have to explain this to every contributor, it will make contributions more complicated and even more sparse :worried:

Flohack74:
@doniks yes its generic, but idk how we can make it more widespread.

doniks:
cleaning up old open PR's ... in case you want to refloat this feel free to reopen!

PR #546, opened 2023-01-22
Update Setting_up.rst

ZGStuff:
EVEN MORE UPDATING VERSION NUMBERS... just update the docs please

### DIFF
--- a/porting/introduction/Setting_up.rst
+++ b/porting/introduction/Setting_up.rst
@@ -15,7 +15,7 @@ When setting up the build environment you need to have Python 3.6 or newer insta
 .. Note::
     Any Linux distribution can easily be set up to switch between Python versions. Consult the documentation for your distribution to find out how this can be done.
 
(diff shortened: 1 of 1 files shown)


doniks:
Can you elaborate a bit on how you tested this/ what the idea is behind these updates? Current version says that there is a difference between 18.04 and 20.04. With your PR you say there is no difference. Is that true? did you retest with both? is 22.04 really different from 20.04? how did you test that? ..

ZGStuff:
mostly just rebuilding ubuntu touch but 22.04... heh
it kinda sucked
a lot of stuff broke

doniks:
Uhm. So, you are basically saying that halium porting on 22.04 does NOT work?! I guess then we should NOT promote. Not at this time. Or am I misunderstanding something?!

doniks:
ok. so I assume then that this is no longer needed. feel free to reopen if you see it differently

PR #561, opened 2023-04-14
replace anbox with waydroid

Fuseteam:

PR #563, opened 2023-05-03
Depreciate Anbox

ElectrodeYT:
Adds a depreciation notice to the Anbox .rst file.

I know the other PR redirects anbox to a in-progress Waydroid doc, however it may still be a good idea to keep the Anbox documentation files, as there may be edge cases (on some older UT devices?) where keeping docs up for Anbox may be a good idea.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -48,7 +48,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'UBports'
-copyright = u'2019-2022, The UBports project (Creative Commons BY-SA 4.0)'
+copyright = u'2019-2023, The UBports project (Creative Commons BY-SA 4.0)'
 author = u'UBports project'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/userguide/dailyuse/anbox.rst
+++ b/userguide/dailyuse/anbox.rst
@@ -1,7 +1,7 @@
 Android apps
 ========================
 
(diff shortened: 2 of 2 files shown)


doniks:
I'm hesitant to keep historic documentation around. There isn't a lot of contributions to even maintain the stuff that is still current and I think it will be worse for out of date information. Git doesn't forget. I admit this is not very accessible, but if people want to pass around links that's always an option.

ElectrodeYT:
I understand. Maybe one could replace the entire anbox.rst file with a warning (and a link to Waydroid) instead? While it would accomplish the same as redirecting, it would make it more obvious that documentation once existed. This would also serve to make people aware that Anbox _was_ deprecated, rather than maybe wonder why they got linked to Waydroid instead, and try to find other Anbox documentation instead.

ElectrodeYT:
> So, my preference would be #561 seeing that anbox upstream basically deprecated it

I wasn't going to try and "replace" 561, now that I think about it, I should have probably left a comment there instead of this...

doniks:
fwiw, there is a redirect if someone has an old anbox link 
https://ci.ubports.com/job/docs.ubports.com/job/PR-561/5/artifact/_build/html/userguide/dailyuse/anbox.html

https://github.com/ubports/docs.ubports.com/pull/561/files#diff-16a7306e360a4be94a15010b4f83525862b470053920dc7436ecb9532a019c4cR20

ElectrodeYT:
Closing as this is better in the discussion for the Waydroid docs PR.

PR #565, opened 2023-05-04
Add documentation on how to port apps to Focal

ElectrodeYT:


doniks (approved):

doniks:
Thanks @ElectrodeYT for this! 

I think it would be good to adjust the first section to give a bit more context. In two years from now maybe nobody will really understand what a focal is :) I think it would be good to mention xenial the previous version and the corresponding version numbers with the names and also add some (rough) dates for context. 

I merged it already, since it was sitting for a long time and it is already a great improvement as is. But if you feel adding another pr to explain the context a bit more, that would be great.

ElectrodeYT:
> Thanks @ElectrodeYT for this!
> 
> I think it would be good to adjust the first section to give a bit more context. In two years from now maybe nobody will really understand what a focal is :) I think it would be good to mention xenial the previous version and the corresponding version numbers with the names and also add some (rough) dates for context.
> 
> I merged it already, since it was sitting for a long time and it is already a great improvement as is. But if you feel adding another pr to explain the context a bit more, that would be great.

I'll do this soon, im also a bit busy with life and other projects :)

Thanks for merging this!

PR #574, opened 2023-09-20
local_testing: arm64 chroot setup commands fixes

beywolfff:
2 small fixes for the chroot setup for arm64

### DIFF
--- a/systemdev/testing-locally.rst
+++ b/systemdev/testing-locally.rst
@@ -109,7 +109,7 @@ Create the directory ``~/logs`` if it does not exist, yet::
 
 Creating a build chroot
 ^^^^^^^^^^^^^^^^^^^^^^^
-In order to create a chroot based on Ubuntu 20.04 (Focal Fossa) with the amd64 architecture under the directory ``/srv/chroot/ubports-${distro}-amd64`` (``chroot_base`` can be changed if needed) the following variables can be defined for later use by the actual commands::
+In order to create a chroot based on Ubuntu 20.04 (Focal Fossa) with the amd64 architecture under the directory ``/srv/chroot/ubports-${chroot_distro}-amd64`` (``chroot_base`` can be changed if needed) the following variables can be defined for later use by the actual commands::
 
     chroot_distro=focal
     chroot_base=/srv/chroot/ubports-${chroot_distro}-amd64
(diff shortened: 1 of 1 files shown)


doniks:
Hey @beywolfff . I must say that the changes are not self evident to me. Can you elaborate a bit why you propose these three changes? (var name, parameter, url)

beywolfff:
Yes ofcouse .

The first one `chroot_distro ` is for consistency with the amd64 commands above and also in the command bellow one of the augment is `chroot_distro ` so if its left `distro` it wont take into account which distribution the user is trying to set up with that command.

The second change `--foreign` this option should be set when the chroot is for cross compilation, example setting up arm64 chroot on amd64 or x86 host machine.  
Source for that information -> https://manpages.debian.org/unstable/sbuild/sbuild-createchroot.8.en.html#foreign

the third  one i believe is a typo when copy/paste http://archive.ubuntu.com/ubuntu/ does not hold arm64 packages this packages can be found in the ports webpage http://ports.ubuntu.com/ubuntu-ports/

systemdev/testing-locally.rst
```diff
@@ -127,8 +127,8 @@ In both cases the chroot will be created by running the following command::
 
 A chroot for cross-building arm64 packages on an amd64 host can e.g. be created under the directory ``/srv/chroot/ubports-${distro}-arm64`` using::
```
maciek134, line 128: The var name needs to be fixed too for consistency

beywolfff: Ahh yes, I missed that , I will change it tonight and update the PR, thank you for noticing.

doniks:
Ah ok. Thanks now I understand. And I agree. With the caveat that, as @maciek134 mentioned there are other places where the `distro` variable name is used. I counted 8 I think. Can you check the whole page and if needed fix/ make consistent?

beywolfff:
> Ah ok. Thanks now I understand. And I agree. With the caveat that, as @maciek134 mentioned there are other places where the `distro` variable name is used. I counted 8 I think. Can you check the whole page and if needed fix/ make consistent?

Yes , I found also that some additional commands must be executed for the cross architecture setup. I will add them too and review the whole page, then I will update this comit.

 Or if you prefer I can cancel this PR and make a new one when I am ready with the changes (aim is for tomorrow to be ready)

doniks:
Awesome. Whatever works best for you! 🙏

beywolfff:
> Awesome. Whatever works best for you! 🙏

I reviewed the page and changed `distro ` to `chroot_distro ` where it was appropriate. I didn't change it in the `crossbuilder` section because its totally different section and way to build and test UT modules so the prefix `chroot_` is irrelevant in that section in my opinion. 

Also I made the commit comment a bit more explanatory for why this changes are done. I believe it is better now.

I am currently testing the sbuild chroot setup and there few commands which should be added for cross-architecture setup , but i am not ready with that yet and i believe it would be better to put that changes in a different PR.

maciek134 (approved):
lgtm

PR #581, opened 2023-10-19
fix readthedocs build

doniks:
rtd has changed and we need to follow. 

I've created a temp build for this branch to confirm the fix:

https://docs.ubports.com/en/fixrtd/systemdev/testing-locally.html

How I've tested by eyeballing:
- that the build looks generally ok, same theme as the current one: https://docs.ubports.com/en/latest/systemdev/testing-locally.html
- has the recent changes on system dev


doniks:
oh cool. now jenkins is crapping out .... hang on

jonnius (approved):
Looks good to me

PR #585, opened 2023-12-15
userguide, porting: update for ADB authorization prompt

peat-psuwit:


userguide/advanceduse/adb.rst
```diff
@@ -39,6 +39,23 @@ Next, you'll need to turn on Developer Mode.
...
+------------------------------
+
+When using adb for the first time on a computer, the following dialog will appear when you unlock the screen.
+
+.. figure:: /_static/images/userguide/adb_authorization_prompt.png
+
+This ensures that ADB commands will not work on a new computer unless user unlock the phone and acknowledge the dialog. Selecting "Allow" will make the ADB commands work for this session, and will also remember this computer so that it won't prompt you again in the future.
+
+.. note::
+    If you're bringing up a port and requires ADB access while the UI is not available yet, you can disable this behavior by editing ``/etc/default/adbd`` and change ``ADBD_SECURE=1`` to ``ADBD_SECURE=0``.
```
doniks, line 54: I think "and requires" is not proper English... It requires, or better you require. Maybe it can also be said a bit more simply. Consider : 



If you're bringing up a port and you require ADB access before the UI is available, you can disable this protection by editing

doniks (changes requested):

userguide/advanceduse/adb.rst
peat-psuwit: Update as suggested. I think "... and require ..." (no 's', sorry) would be as valid as "... and you require", but I agree that your whole sentence is a bit more tidy.

doniks (approved):

Issue #588, opened 2024-01-25
Docs mention Lubuntu Software Centre which does not exist


Caedorus:
See title. I've been told it's been depreciated.

PR #589, opened 2024-01-25
Remove reference to Lubuntu Software Centre

Caedorus:
Lubuntu Software Centre is depreciated, therefore no longer in the repos and cannot be installed. Changed it to point to gnome-software instead.

### DIFF
--- a/userguide/dailyuse/libertine.rst
+++ b/userguide/dailyuse/libertine.rst
@@ -72,7 +72,7 @@ From the apps list in the *Manage Libertine Containers* menu you can tap the + s
 
 Enter the package name if you know it.
 Otherwise you can search the archives for a package.
-You can make this process easier by making the *Lubuntu Software Center* your first package install.
+You can make this process easier by making a software center like gnome-software your first package install.
 Being a desktop app, you will need to use the *Libertine Tweak Tool* in the *Open Store* to make the text large enough to read on a mobile device.
 
 Packages may also be installed via the command-line::


Caedorus:
Fixes #588

userguide/dailyuse/libertine.rst
```diff
@@ -72,7 +72,7 @@ From the apps list in the *Manage Libertine Containers* menu you can tap the + s
 
 Enter the package name if you know it.
 Otherwise you can search the archives for a package.
-You can make this process easier by making the *Lubuntu Software Center* your first package install.
+You can make this process easier by making a software center like gnome-software your first package install.
```
doniks, line 75: I'm wondering ... did you try to install gnome-software? is it possible? does it run? does it run well enough to justify the claim that this makes the "process easier"?

It's been a long long time that I looked at this, but I think gnome-software is a rather .... whatstheword .... advanced ... sophisticated ... modern application. I'm wondering how well it actually runs in libertine. did you test it? can you comment?

Caedorus: I did test it, it worked fine though I would recommend attaching a mouse for ease of use (I can't get the scaling to work yet)

Caedorus: Alternatively feel free to suggest another one, the one to change it *to* isn't really the point of this PR

doniks: > I did test it, (fp4 focal ota-4) it works fine though I would recommend attaching a mouse for ease of use (I can't get the scaling to work yet)

great! thanks

doniks (approved):

PR #594, opened 2024-02-13
Update to the latest version of Sphinx

ElectrodeYT:
I have not been able to find any (bad) differences between the current live version and when using the latest version of Sphinx/Furo. As far as I can tell, the stability policy from Furo ([available here](https://pradyunsg.me/furo/stability/)), keeping both at simply the latest version available should be safe.

The ToC change is needed for the index page because it seems to generate an empty Contents list for some reason. Given the fact that the Sphinx 4.5.0 generated page has no contents bar here, hiding it here seems like the correct call.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,6 +13,8 @@ pipeline {
     stages {
         stage("Install dependencies") {
             steps {
+                sh 'python --version'
+                sh 'pip --version'
                 // TODO: the `--quiet` is to workaround some Docker/Jenkins bugs/misconfigurations, this should be removed at some point
                 sh 'pip install --quiet --no-cache-dir --user --upgrade -r requirements.txt'
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
     agent {
         docker {
-            image "python:3.7"
+            image "python:3.12"
         }
     }
     options {
@@ -13,7 +13,8 @@ pipeline {
     stages {
         stage("Install dependencies") {
             steps {
-                sh 'pip install --no-cache-dir --user --upgrade -r requirements.txt'
+                // TODO: the `--quiet` is to workaround some Docker/Jenkins bugs/misconfigurations, this should be removed at some point
+                sh 'pip install --quiet --no-cache-dir --user --upgrade -r requirements.txt'
             }
         }
         stage("Build docs"){
--- a/index.rst
+++ b/index.rst
@@ -1,3 +1,5 @@
+:hide-toc:
+
 Welcome to the UBports documentation!
 =====================================
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-sphinx==4.5.0
-furo==2021.6.24b37
+sphinx==7.2.6
+furo==2024.1.29
 sphinx-intl
 sphinxext-rediraffe


ElectrodeYT:
From what I can tell, the CI Failure is likely due to the PR being untrusted, according to Jenkins. I am not sure if my Jenkinsfile changes would make CI work, but I am also not sure how to verify this either.

doniks:
> From what I can tell, the CI Failure is likely due to the PR being untrusted, according to Jenkins. I am not sure if my Jenkinsfile changes would make CI work, but I am also not sure how to verify this either.

here https://ci.ubports.com/blue/organizations/jenkins/docs.ubports.com/detail/PR-594/8/pipeline it says something about sphinx version 7.2.6 not being available

doniks:
Not sure what determines these versions ... looks like up to 5.3.0 would be available. 

To recap: You're proposal is meant to fix the currently broken local build of master. `./build.sh` (After removing any old copy of the python environment in ~/ubportsdocsenv) results in the following error:

```
15:34 $ ./build.sh 
Build environment found in "/home/peter/ubportsdocsenv".
Building...
Running Sphinx v4.5.0
loading translations [en]... done

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
```

so, I can think of two options:
* you could try with 5.3.0 and see whether sphinx,furo,jenkins, everybody is happy with that
* we try to chase someone from Infrastructure team with Jenkins access why sphinx can't go higher

ElectrodeYT:
From my small amount of research, the 5.3.0 Sphinx problem is probably related to the python version the Jenkins container currently uses.

As can be seen on PyPI, [Sphinx 5.3.0](https://pypi.org/project/Sphinx/5.3.0/) needs python 3.6, while [Sphinx 6.0.0](https://pypi.org/project/Sphinx/6.0.0/) needs python 3.8, and the current Jenkinsfile is python 3.7.

ElectrodeYT:
I think the python container version is the cause here.

`python:3.7` docker container:
```
^[[A[alexander@alex-computer docs.ubports.com]$ sudo docker run --rm -it --entrypoint /bin/bash python:3.7
root@8eda44e65742:/# pip3 install sphinx
Collecting sphinx
  Downloading sphinx-5.3.0-py3-none-any.whl (3.2 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3.2/3.2 MB 4.7 MB/s eta 0:00:00
```

`python:latest` docker container:
```
[alexander@alex-computer docs.ubports.com]$ sudo docker run --rm -it --entrypoint /bin/bash python:latest
root@91dc3995efe5:/# pip3 install sphinx
Collecting sphinx
  Downloading sphinx-7.2.6-py3-none-any.whl.metadata (5.9 kB)
```

ElectrodeYT:
I have updated the Jenkinsfile change to set Jenkins to use the `python:3.12` container instead of `python:latest`. This should fix CI; I do not know, however, how to make Jenkins use the Jenkinsfile changes for CI on this PR.

doniks:
so, 

* we can pick the python version via Jenkins file, see here: https://ci.ubports.com/blue/organizations/jenkins/docs.ubports.com/detail/PR-596/7/pipeline#step-24-log-2
* however, I cannot get the whole package to build ...  python 3.12 + sphinx 7.2.6 + furo 2024.1.29 still fails : https://ci.ubports.com/blue/organizations/jenkins/docs.ubports.com/detail/PR-596/7/pipeline/#step-25-log-32

@ElectrodeYT can you tell again which combination of versions of python + sphinx + furo you could build it with?

doniks:
argggghhhh Jenkins! It still fails to build. while it's literally the same change as that: https://github.com/ubports/docs.ubports.com/pull/596/files so maybe there is a difference between *internal* branches created inside the ubports/docs.ubports.com repo versus external ones such as from a user fork like yours .... when I look here https://ci.ubports.com/blue/organizations/jenkins/docs.ubports.com/detail/PR-594/11/pipeline the python and pip versions do not get printed. Maybe the easiest is to just proceed with that PR https://github.com/ubports/docs.ubports.com/pull/596/files. I shall add your comment about --quiet just being a hacky workaround ... still need to properly check the output also

ElectrodeYT:
Yep, thats Jenkins; iirc the thing it does is not use Jenkinsfile changes from "untrusted" PRs.

doniks:
always another way : https://github.com/ubports/docs.ubports.com/pull/597

doniks:
ok, so for the protocol, there are some known issues with this update, but they are not show stopping and we also need to move ahead with the version upgrade, otherwise other contributions are blocked. maybe once it's live and has more eyeballs on it someone will find ways to further improve this

doniks (approved):

PR #597, opened 2024-02-22
Sphinx 7 update

doniks:
I've pushed the branch from PR https://github.com/ubports/docs.ubports.com/pull/594 into this internal branch to make sure Jenkins actually uses the new configuration.

To recap:
- current master of the docs can't be built locally anymore because it's lagging behind in updates
- this PR lifts up versions to make it able to build
- open question is: are there regressions with this new build - see discussion below

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,6 +13,8 @@ pipeline {
     stages {
         stage("Install dependencies") {
             steps {
+                sh 'python --version'
+                sh 'pip --version'
                 // TODO: the `--quiet` is to workaround some Docker/Jenkins bugs/misconfigurations, this should be removed at some point
                 sh 'pip install --quiet --no-cache-dir --user --upgrade -r requirements.txt'
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
     agent {
         docker {
-            image "python:3.7"
+            image "python:3.12"
         }
     }
     options {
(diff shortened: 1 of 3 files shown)


doniks:
I do see some differences:

* left: PR
* right: current public version

Issues:
* empty space on top
* giant "next" arrow
* the "foldout menu from the left" has some strange bright shadow
* in fact, that shadow is also there when I fold out the menu from the right

Those are not nice

Further differences, that are really not an issue for me, but for the protocol
* links are blue ... in fact ... *I* prefer it, but I'm very much not a graphics guy
* "Notes" look a bit different
* I thiiiink the fonts of headlines are slightly bigger
* colors especially in light theme looks slightly .... less subtle and nice in the PR

image
image
image
image
image

I think it would be good if we could at least fix those four "Issues" ... @ElectrodeYT is this in your comfort zone? Are you willing to look into this? If not maybe we can shop around for other help ... I mean worst case, I'd also just roll with it, if we really have to, but ideally it could be a bit more polished

ElectrodeYT:
Huh, other than the arrow none of these issues were visible to me in the local HTML version. Will look into it

ElectrodeYT:
I have looked a bit more into this, and determined the following:

- Can't instantly find a (supported by furo) way to remove the arrows
- The theming issues (the gap on top, wrong colours for text, size and fonts...) seem to be caused by how Jenkins seems to host the artifact files; downloading the build and opening it locally works. (Jenkins being weird is becoming a theme here...)
- Not sure if this is a local build thing, a firefox scaling thing, or a actual problem, but the text does seem to be a bit bigger
- The chapter titles in the side bar have a brighter colour, at least on the dark theme

doniks:
mhm. okay. too bad, but fine, let's roll with it then. I'll go ahead and approve and merge https://github.com/ubports/docs.ubports.com/pull/594 and close my draft #597 here

doniks:
ugh ... damn ... all those issues apparently where jenkins related. I merged your PR @ElectrodeYT and it all looks fine. sorry I wasted your time :shrug:

PR #601, opened 2024-04-14
Update standalone_kernel_build.rst

Umeaboy:
Added Mageia dependencies.


doniks (changes requested):
Please check the formatting. It should be a code block https://ci.ubports.com/job/docs.ubports.com/job/PR-601/1/artifact/_build/html/porting/build_and_boot/standalone_kernel_build.html I think that's down to the number of colons `:` and/or newlines and/or indentation 

Oh and while you're at it, maybe you could double check the apt block above as well. It seems to have some useless backslashes `\`

doniks (approved):

PR #602, opened 2024-04-21
Fix typos

birdoplank:


Makefile
```diff
@@ -1,6 +1,7 @@
 # Minimal makefile for Sphinx documentation
 #
 
+
```
jonnius, line 4: I don't see which typo this is supposed to fix. I only see white space changes. Can you please elaborate?

jonnius:
Looking at the profile and all the other alike PRs the author made, I can only consider this spam.

Makefile
Janmm14: The user in question has first tried to introduce curl script download into the build process in other repos and is later jsut spamming mostly empty line additons to various repositories

PR #607, opened 2024-09-18
drop the legacy android devices section

Fuseteam:
these devices are no longer supported on focal due to the kernel being too old and the kernel source being lost in some cases


doniks:
Looks good to me. I'm not merging it yet, to see whether someone still wants to chime in on this archived docs discussion (ps I've branched current master already into a branch called xenial)

doniks (approved):

doniks:
Seems there's consensus in the telegram chat. I'm merging

PR #609, opened 2024-09-19
update ssh documentation for focal

Fuseteam:
we now use systemd on focal, this opens a lot of possibilties with ssh

i've documented my current findings

### DIFF
--- a/userguide/advanceduse/ssh.rst
+++ b/userguide/advanceduse/ssh.rst
@@ -37,7 +37,7 @@ Let's assume it's stored as ``/home/phablet/id_rsa.pub``. Use the terminal app o
 
 Now start the SSH server::
 
-    sudo service ssh start
+    sudo systemctl start ssh.socket
 
 
 Connect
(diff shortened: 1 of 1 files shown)


userguide/advanceduse/ssh.rst
```diff
@@ -70,6 +70,23 @@ In this case you use the second IP address
...
+    sudo systemctl enable ssh.socket
+
+you can edit the port ssh is listening on with the following command::
+    
+    sudo systemctl edit ssh.socket
+
+then add the following text::
+
+    [Socket]
+    ListenStream=
```
doniks, line 87: What does this line do? Is it needed?

doniks (changes requested):
Pls check my comment

userguide/advanceduse/ssh.rst
Fuseteam: yes this line reset the socker it is listening one, without this line it would listen on both 22 and whatever the user sets

doniks (approved):

PR #612, opened 2024-09-27
Update reverse-tethering.rst

Fuseteam:
the steps to enable rndis has changed in focal, it is now done via gdbus

### DIFF
--- a/userguide/advanceduse/reverse-tethering.rst
+++ b/userguide/advanceduse/reverse-tethering.rst
@@ -54,9 +54,9 @@ Steps
 #. If it's not working yet, on your *computer*:
 
    - Turn on IP forwarding: ``echo 1 | sudo tee /proc/sys/net/ipv4/ip_forward``
-   - Apply NAT: ``sudo iptables -t nat -A POSTROUTING -s 10.42.0.0/24 -o eth0 -j MASQUERADE``
-
-     - Make sure you replace ``eth0`` by the interface that connects your computer to the internet.
+   - Apply Forwarding iptables Rule: ``sudo iptables -P FORWARD ACCEPT``
+   - Apply NAT rule: ``sudo iptables -t nat -A POSTROUTING -s 10.42.0.0/24 -j MASQUERADE``
+   - reconfigure the network interface: ``sudo ifconfig 10.42.0.149 netmask 255.255.255.0``
 
    - NOTE: These steps are usually performed automatically by the NetworkManager on Ubuntu 22.04.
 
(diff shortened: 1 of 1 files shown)


Fuseteam:
granted i can't seem to get this working. seems the traffic goes to my pc, but no further

doniks:
> granted i can't seem to get this working. seems the traffic goes to my pc, but no further

So, are you looking further into this, should we wait? Or ready to merge? I'm fine either way, up to you

doniks:
Thinking again, if you can't make an internet connection, just local traffic, that's kind of a big deal, so at least this should be documented.

Also, I see now the two "Known Issues" - can you say anything about those?

Fuseteam:
known issues? which ones?
i can ping the pc from there but that's normal, i can confirm it at least get routed to the pc (so traffice within 10.42.0.x works)

Fuseteam:
i would prefer if the actual reverse tether works, but if your fine with merging in the current state that's fine too. I don't particularly see why it doesn't get routed through. and according to the notes network manager is supposed to handle the pc side......

Fuseteam:
ah those known issues, i cannot say much about those as currently i haven't gotten it working yet

doniks:
Mhm. It's a complicated choice. I'd say

1. (Under the assumption that the current version of the docs just plain does not work anymore) it's better to merge this PR, since that is the best known version working in focal
2. However, since not getting an internet connection is kind of a big deal I'd suggest to put a prominent warning. Maybe a "Note:" between the first paragraph and the "Steps" section saying something like: The current version provides a local network connection between the PC and the UT device. Probably something needs to be improved in the routing to get an actual internet connection.
3. Let's maybe leave the Known Issues as is for now, but they should definitely be retested once we get the internet connection working

Fuseteam:
hmmm given the whole point is to get internet connection working.......
i'll need to ponder how to word the warning, i am beginning to suspect lxc and docker networking interfering. so it might be a me thing. It would nice if someone else could test the steps to see if it works or not

Fuseteam:
i think i found it, if i check `route` on my pc, i see this line
```
default         tomoya          0.0.0.0         UG    101    0        0 enxc2899686591b
```
where `tomoya` is my ut device so i'm guessing that why i set the gateway to be my pc on my phone, my pc has my phone as the gateway. So it doesn't go anywhere

Fuseteam:
`developer_mode` seems to not set any gateway but still no dice grrr

Fuseteam:
got it :partying_face: i found these steps under "Bridging host PC to allow device to reach the Internet" on https://developer.ridgerun.com/wiki/index.php/How_to_use_USB_device_networking fixed it so i added in the docs (ironically it is linked on the docs too xd)

Fuseteam:
i guess the only thing left now is, that this whole gdbus command has a ui setting in ut tweak tool while [a UI settings is in the works](https://gitlab.com/ubports/development/core/lomiri-system-settings/-/merge_requests/437)
 should add either of those options instead?

PR #614, opened 2024-10-05
Update tethering docs

Fuseteam:
i'm annoying but i messed with reverse tethering a bit more and it turns out it's actually much simpler

in usb tethering mode we get a network interface using "the shared ipv4 method" on UT and an network interface using "the automatic ipv4 method"

the automatic method is familiar, this is where we automatically get assigned an ip address and dns is setup automatically

the shared method does the following:
- enables IP forwarding for the interface;
- adds firewall rules and enables masquerading;
- starts dnsmasq as a DHCP and DNS server.

in my testing what it doesn't do is `iptables -P FORWARD ACCEPT` 

so for reverse tethering we want the interface on the computer to use the shared method and the interface on the phone to use the automatic method, so i rewrote the steps to reflect that. now we no longer need to fuss with ip addresses, gateways and dns servers

i've yet to confirm if "enable IP forwarding for the interface" makes `echo 1 | sudo tee /proc/sys/net/ipv4/ip_forward` obsolete, it seems that `cat /proc/sys/net/ipv4/ip_forward` already returned `1` for me

### DIFF
--- a/userguide/advanceduse/reverse-tethering.rst
+++ b/userguide/advanceduse/reverse-tethering.rst
@@ -9,17 +9,17 @@ Steps to Set Up Reverse Tethering
 
 #. Prepare your device (Ubuntu Touch) and your computer:
 
-   - Connect your computer to the internet.
-   - Attach your device to your computer via USB.
+   - Connect your *computer* to the internet.
+   - Attach your *device* to your *computer* via USB.
+   - Turn off Wifi and Data on your *device*
 
-#. On the *device*:
+#. Run the following commands on your *device*:
 
-   - Turn off Wifi and Data to avoid the gateways from clashing
    - Set your device usb in tethering mode: ``gdbus call --system --dest com.meego.usb_moded --object-path /com/meego/usb_moded --method com.meego.usb_moded.set_mode rndis_adb``
    - Bring your tethering connection down: ``sudo nmcli connection down tethering``
    - Modify your tethering connection: ``sudo nmcli connection modify tethering ipv4.method auto``
 
-#. On your *computer*: 
+#. Run the following commands on your *computer*: 
 
    - Get your interface name: ``ip route show to default via 10.42.0.1``
      ::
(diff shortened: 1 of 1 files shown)


Fuseteam:
i have now created [a script](https://github.com/Fuseteam/tethering-nm) that can be run on the pc side to simplify it xD

userguide/advanceduse/reverse-tethering.rst
```diff
@@ -15,60 +15,61 @@ Steps
...
+       WIRED-PROPERTIES.CARRIER:               on
+       IP4.ADDRESS[1]:                         10.42.0.118/24
+       IP4.GATEWAY:                            10.42.0.1
+       IP4.ROUTE[1]:                           dst = 10.42.0.0/24, nh = 0.0.0.0, mt = 101
+       IP4.ROUTE[2]:                           dst = 0.0.0.0/0, nh = 10.42.0.1, mt = 101
+       IP4.DNS[1]:                             10.42.0.1
+       IP6.ADDRESS[1]:                         fe80::692:2a1:8127:91a1/64
+       IP6.GATEWAY:                            --
+       IP6.ROUTE[1]:                           dst = fe80::/64, nh = ::, mt = 1024
+
```
doniks, line 50: this is quite hard to read. How about 

nmcli --fields GENERAL.TYPE device show enx122626d0fe26

userguide/advanceduse/reverse-tethering.rst
```diff
@@ -15,60 +15,61 @@ Steps
...
+   - Your connection name is ``Wired connection 2``
+   - Bring the connection down: Run ``nmcli connection down Wired\ connection\ 2``
```
doniks, line 52: Consider quotes instead of backslashes to make it a little more human readble

nmcli connection down 'Wired connection 2'

doniks (changes requested):
Are all these steps needed for everyone? Or is part of this more troubleshooting steps that may not be needed for everyone?!

Could it be reorganised similar to this:

# Enable reverse tethering
1. connect cable
2. on dev
  * turn off wifi and data
  * run gdbus

Now you should have a working internet connection.

# Troubleshooting

If this doesn't give you a working internet connection yet, you might ........

doniks:
Also I find the style of writing with the colons and "Run" a bit clunky

* do something: Run `command`

I think this could just be

* do something: `command`

Except maybe for the first command we should say run this on your computer/device or something like that, but the repeated ": Run" look odd to me

userguide/advanceduse/reverse-tethering.rst
Fuseteam: oh that might work, lemme check that

Fuseteam:
yeah fair i'll change that

but yes changing the ipv4 method is required for everyone, should we not want to mess with ip addresses, gateways and dnses

userguide/advanceduse/reverse-tethering.rst
Fuseteam: yes i agree, i'll do that :D

userguide/advanceduse/reverse-tethering.rst
Fuseteam: i was looking for a way to make the output smaller, thanks!

Fuseteam:
i guess the only way to simplify it further would be if we can get usb-moded to setup a network interface using the auto or manual method :D

doniks (approved):

PR #615, opened 2024-10-05
Set expectations for non-native app packages

capsia37:
Set expectations for 'non-native' app packages to avoid common misconceptions

### DIFF
--- a/userguide/dailyuse/libertine.rst
+++ b/userguide/dailyuse/libertine.rst
@@ -10,7 +10,7 @@ Limitations
 
 Before using Libertine on your device, be aware that there are limitations to the usability of packages.
 
-- Software that is not designed with a mobile use-case in mind will be difficult to use due to lack of touchscreen support, scaling and other UI issues.
+- Software that is not designed with a mobile use-case in mind may be difficult to use due to lack of touchscreen support, scaling and other UI issues.
 - Applications will not run in the background in Libertine, hence it is not a solution to run server software on your phone.
 
 Manage containers
--- a/userguide/dailyuse/libertine.rst
+++ b/userguide/dailyuse/libertine.rst
@@ -5,6 +5,14 @@ Libertine allows you to use standard desktop applications in Ubuntu Touch.
 
 To install applications you can use the *Settings* interface on your Ubuntu Touch device, or the command-line as described further below.
 
+Limitations
+-----------
+
+Before using Libertine on your device, be aware that there are limitations to the usability of packages.
+
+- Software that is not designed with a mobile use-case in mind will be difficult to use due to lack of touchscreen support, scaling and other UI issues.
+- Applications will not run in the background in Libertine, hence it is not a solution to run server software on your phone.
+
 Manage containers
 -----------------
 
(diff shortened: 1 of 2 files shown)


userguide/dailyuse/libertine.rst
```diff
@@ -5,6 +5,14 @@ Libertine allows you to use standard desktop applications in Ubuntu Touch.
 
 To install applications you can use the *Settings* interface on your Ubuntu Touch device, or the command-line as described further below.
 
+Limitations
+-----------
+
+Before using Libertine on your device, be aware that there are limitations to the usability of packages.
+
+- Software that is not designed with a mobile use-case in mind will be difficult to use due to lack of touchscreen support, scaling and other UI issues.
```
doniks, line 13: consider: "will" -> "may"

userguide/dailyuse/libertine.rst
```diff
@@ -5,6 +5,14 @@ Libertine allows you to use standard desktop applications in Ubuntu Touch.
...
+- Applications will not run in the background in Libertine, hence it is not a solution to run server software on your phone.
```
doniks, line 14: Consider: "Applications will not run automatically in the background in Libertine, hence it is not an out of the box solution to run server software on your phone."

userguide/dailyuse/waydroid.rst
```diff
@@ -6,6 +6,14 @@ Android apps
 .. note::
     You will need to execute commands on your Ubuntu Touch device (locally via the terminal application or remotely via :doc:`ADB </userguide/advanceduse/adb>`/:doc:`SSH </userguide/advanceduse/ssh>` ) to initialize Waydroid, install APKs and uninstall applications.
 
+Limitations
+-----------
+
+Before installing Waydroid on your device, please note that there are limitations to its support and integration:
+
+- After installing Waydroid on Ubuntu Touch, you may notice higher battery consumption. As a workaround, Waydroid can be stopped manually when not in use.
```
doniks, line 14: consider "As a workaround" -> "To mitigate this"

userguide/dailyuse/waydroid.rst
```diff
@@ -6,6 +6,14 @@ Android apps
...
+- Not all Android applications will work with Waydroid. Apps that require Google's Play Integrity service or SafetyNet, such as banking apps, are unlikely to run.
+- Problems may occur if both Ubuntu Touch and Waydroid request the use of the same peripherals, and one will not be able to claim them.
```
doniks, line 16: If you know of an example it might be nice to add it

doniks (changes requested):

userguide/dailyuse/waydroid.rst
capsia37: Added camera access issue as example: https://gitlab.com/ubports/porting/reference-device-ports/android11/fairphone-4/fairphone-fp4/-/issues/67

userguide/dailyuse/libertine.rst
capsia37: I'm not sure about this one. Afaik, that's not a supported use case, so we shouldn't promote it even if it could be done. 'will not run automatically in the background' sounds rather possibilistic about having a manual procedure to do it the supported way.

doniks: ah, *supported* .... yeah, that I don't know. I was more thinking along the lines of technically *possible*. It doesn't seem so far fetched to write a systemd file in the home dir starting some executable that happens to reside inside the libertine directory. ok. lets leave it as is

doniks (approved):

PR #621, opened 2024-11-08
update zh_CN locales files by `msgmerge`

kuailexs:
after https://github.com/ubports/docs.ubports.com/pull/619


doniks:
Shouldn't this happen via weblate?

kuailexs:
> Shouldn't this happen via weblate?

this repository on weblate is closed

kuailexs:
https://hosted.weblate.org/projects/ubports/

only  docs.ubports.com about page

kuailexs:
There is no place to contribute translations

doniks:
ah. yes. I see now. you're right. Thanks for pushing this even with lacking infrastructure!

doniks (approved):

PR #632, opened 2024-12-19
refactor: MAJOR refactor/beautify of build.sh

grenudi:
IMPORTANT note:
I removed the ability to pass UBPORTSDOCSENV straight after ./build.sh
like so:
```sh
./build.sh ~/ubportsdocsenv/ 
```
now it's:
```sh
./build.sh --env ~/ubportsdocsenv/ 
```
or for the full output like before it's now:
```sh
./build.sh --env ~/ubportsdocsenv/ --verbose
```

image

image

image

image

image

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,53 +2,200 @@
 
 set -Eeou pipefail
 
-# color codes
+# fold me :) (in vscode vscodium etc)
+#region DEFAULTS_AND_COMMAND_LINE_ARGUMENTS
+# Parse command-line arguments for flags
+verbose=false
+UBPORTSDOCSENV=${UBPORTSDOCSENV:-"$HOME/ubportsdocsenv"}
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    --verbose)
+      verbose=true
+      shift
+      ;;
+    --env)
+      # If --env flag is passed, set UBPORTSDOCSENV to the next argument value
+      UBPORTSDOCSENV=${2-$UBPORTSDOCSENV}
+      shift 2
+      ;;
+    *)
+      shift
+      ;;
+  esac
+done
+#endregion DEFAULTS_AND_COMMAND_LINE_ARGUMENTS
+
+#region VARIABLES & FUNCTIONS
+#region COLOR_CODES
+#region COLOR_CODES_CONSTANTS
 GREEN='\033[0;32m'
 YELLOW='\033[1;33m'
 RED='\033[0;31m'
 PLAIN='\033[0m'
 
-# build directory
-## use environment variable if it is set, or default value
-UBPORTSDOCSENV=${UBPORTSDOCSENV-"$HOME/ubportsdocsenv"}
-## use commandline parameter if it is given
-UBPORTSDOCSENV=${1-$UBPORTSDOCSENV}
+SUCCESS=$GREEN
+WARNING=$YELLOW
+ERROR=$RED
+PLAIN_TEXT=$PLAIN
+#endregion COLOR_CODES_CONSTANTS
 
-# check if virtualenv already exists
-if [ -d "$UBPORTSDOCSENV" ]; then
-  echo -e "${GREEN}Activating virtual build environment found in \"$UBPORTSDOCSENV\".${PLAIN}"
-  source $UBPORTSDOCSENV/bin/activate
-else
-  echo -e "${YELLOW}Build environment not found in \"$UBPORTSDOCSENV\" ... creating it.${PLAIN}"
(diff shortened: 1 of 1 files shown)


grenudi:
i think i need to find a better use of my spare time x`) ...

build.sh
```diff
@@ -2,53 +2,200 @@
...
+      verbose=true
+      shift
+      ;;
+    --env)
+      # If --env flag is passed, set UBPORTSDOCSENV to the next argument value
+      UBPORTSDOCSENV=${2-$UBPORTSDOCSENV}
+      shift 2
+      ;;
+    *)
+      shift
```
doniks, line 22: Let's be a little more defensive here and error out instead of silently ignoring invalid options

build.sh
```diff
@@ -2,53 +2,200 @@
...
+      ;;
+  esac
```
doniks, line 24: shift here instead

build.sh
```diff
@@ -2,53 +2,200 @@
 
 set -Eeou pipefail
 
-# color codes
+# fold me :) (in vscode vscodium etc)
```
doniks, line 5: I don't really see the point of a "fold me" comment. Even more so a reference to one particular editor. I don't think this is helpful 

Frankly I find the region/end region comments more distracting than helpful. Please remove

build.sh
```diff
@@ -2,53 +2,200 @@
...
+            fi
+        else
+            echo_plain "$package is already installed."
+        fi
+    done
+}
+
+
+# Function to install dependencies for most Linux systems
+install_dependencies_for_most_linux_systems() {
```
doniks, line 135: Can we just call this `install_dependencies` please

build.sh
```diff
@@ -2,53 +2,200 @@
...
+  echo_success "Checking dependencies"
```
doniks, line 136: Shouldn't this be plain?

doniks: Maybe the detected xyz could be the success?!

build.sh
```diff
@@ -2,53 +2,200 @@
...
+  rm -rf _build/
+  if [ "$(uname)" == "Linux" ]; then
+    sphinx-build -Wa . _build/html -j `nproc --all` || exit $?
+  else
+    sphinx-build -Wa . _build/html -j `sysctl -n hw.ncpu` || exit $?
+  fi
+  sphinx-build -Wab rediraffecheckdiff . _build/html
+}
+build_documentation(){
+  echo_success "Building..."
```
doniks, line 183: Plain

build.sh
```diff
@@ -2,53 +2,200 @@
...
+
+#region BUILD
+wrapper_build_documentation(){
+  rm -rf _build/
+  if [ "$(uname)" == "Linux" ]; then
+    sphinx-build -Wa . _build/html -j `nproc --all` || exit $?
+  else
+    sphinx-build -Wa . _build/html -j `sysctl -n hw.ncpu` || exit $?
+  fi
+  sphinx-build -Wab rediraffecheckdiff . _build/html
```
doniks, line 180: Should this also `|| exit` ? 

That should also allow you to put a echo_success after

build.sh
```diff
@@ -2,53 +2,200 @@
...
-      echo "Detected apt"
-      sudo apt install python3-pip python3-virtualenv
+      install_via_apt
   elif command -v pacman &> /dev/null; then
-      echo "Detected pacman"
-      sudo pacman -S --noconfirm python-pip python-virtualenv
+      install_via_pacman
   else
-      echo "Unsupported package manager. Please install the packages and create the virtualenv manually."
+      echo_error "Unsupported package manager. Please install the packages and create the virtualenv manually."
```
doniks, line 145: The way it's implemented right now with `exit 1`, it's misleading to say users should install dependecies, since even if they did, it would still error out. Better to be factual and just say "Failed to detect package manager"

If the exit would be removed, the script could continue. In that case this should be a warning about package manager detection. 

In no case does it make sense to instruct about virtualenv, since this would be a separate subsequent step.

doniks (changes requested):

doniks:
@grenudi maybe have a look at update_translations.sh script. Looks like as if it started as a copy/paste of an older build script. By now it's diverged a lot and with your change would diverge more. Maybe this could be added here instead with a --translation option or something

grenudi:
> @grenudi maybe have a look at update_translations.sh script. Looks like as if it started as a copy/paste of an older build script. By now it's diverged a lot and with your change would diverge more. Maybe this could be added here instead with a --translation option or something 

ty, I will look into it

